PR #9, opened 2013-07-01
Align connector detectWeb rules with Firefox

aurimasv:
Chrome connector seems to like this. I'm having a bit of a hard time finding test cases atm (some of the pages that were reported don't seem to have embedded iframes anymore), but reverting https://github.com/zotero/translators/commit/2ca80898871b5039cf25c330859ee195885e11bc and navigating to http://www.forbes.com/sites/peterlipson/2013/04/19/a-film-producer-a-cancer-doctor-and-their-critics/ doesn't work on the current connector, but does with this patch

I haven't tested Safari connector.

### DIFF
--- a/src/common/inject/inject.js
+++ b/src/common/inject/inject.js
@@ -157,7 +157,12 @@ Zotero.Inject = new function() {
 				_translate.setHandler("translators", function(obj, translators) {
 					if(translators.length) {
 						me.translators = translators;
+						translators.targetURL = document.location.href;
+						translators.targetsTopFrame = isTopWindow;
 						for(var i=0; i<translators.length; i++) {
+							translators.bestPriority = translators[0].priority;
+							
+							// Safari has issues cloning functions, so strip them off
 							if(translators[i].properToProxy) {
 								delete translators[i].properToProxy;
 							}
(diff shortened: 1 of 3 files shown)


aurimasv:
I rebased this and reverted the change [discussed above for Safari](https://github.com/zotero/zotero-connectors/pull/9#discussion_r5036673). Did some testing and it seems to work well. This fixes the issue reported [here](https://forums.zotero.org/discussion/38752/financial-times-translator-giving-an-error-message/), which was due to translator attaching to the wrong iframe on Chrome/Safari

src/common/inject/inject.js
```diff
@@ -158,6 +158,11 @@ Zotero.Inject = new function() {
 					if(translators.length) {
 						me.translators = translators;
 						for(var i=0; i<translators.length; i++) {
+							translators.bestPriority = translators[0].priority;
+							translators.targetURL = document.location.href;
+							translators.targetsTopFrame = document.defaultView == document.defaultView.top;
```
simonster: This could throw due to SOP. You can just use `isTopWindow` (defined in this file using `try`/`catch`). Also these don't need to be in a loop.

aurimasv:
Done and squashed

adomasven:
I've referenced this PR before when modifying this code, and the rules are currently aligned. Closing.

PR #30, opened 2014-12-01
Mutation observers

aurimasv:


aurimasv:
bump

adomasven:
I was never sure what this was intended to fix, but from looking at the code and the issues referenced, nothing that we're suffering from anymore.

PR #36, opened 2015-06-04
Return `false` when canceling multiples selection dialog

aurimasv:
Re https://forums.zotero.org/discussion/49763

(edit: nvm, I tested it and this does fix it)

I suppose generally this wouldn't matter. WoS seems to export all results if no result IDs are supplied though.


aurimasv:
Fixed in WoS translator in the mean time https://github.com/zotero/translators/commit/10da65fca47433f8af1c16a6d52c7fbd9b6226a8

PR #51, opened 2016-09-06
Update to support Zotero 5.0 translator code sharing

adomasven:


src/common/zotero.js
```diff
@@ -24,17 +24,25 @@
...
+	// http://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
+	// Firefox 1.0+
+	this.isFirefox = typeof InstallTrigger !== 'undefined';
+	// At least Safari 3+: "[object HTMLElementConstructor]"
+	this.isSafari = Object.prototype.toString.call(window.HTMLElement).indexOf('Constructor') > 0;
+	// Internet Explorer 6-11
+	this.isIE = /*@cc_on!@*/false || !!document.documentMode;
+	// Edge 20+
+	this.isEdge = !this.isIE && !!window.StyleMedia;
+	// Chrome 13+ A deprecated extension function
```
dstillman: Outdated comment?

gulpfile.js
```diff
@@ -0,0 +1,128 @@
...
+});
+
+gulp.task('babelify', babelify);
+gulp.task('minify', minify);
+
+gulp.task('build-dev', sequence('build-sh-dev', 'babelify'));
+
+gulp.task('build', sequence('build-sh', 'babelify', 'minify'));
+
+gulp.task('default', ['build-dev']);
```
dstillman: Spaces -> tabs

gulpfile.js
```diff
@@ -61,14 +62,24 @@ function reloadChromeExtensionsTab(cb) {
...
         .pipe(plumber({errorHandler: onError}))
         .pipe(babel({
             presets: ['es2015']
         }))
-        .pipe(gulp.dest('./build/safari.safariextension/zotero/'));
+        .pipe(gulp.dest((data) => data.base));
+}
+
+function minify() {
+	return gulp.src('./build/**/*.js')
```
dstillman: Only the bookmarklet code gets minified, not connector code. (Can't submit minified code to Mozilla.)

dstillman: These lines also don't work properly if a custom build directory is specified in config.sh. That's OK, but then we should remove that config option and always build to ./build.

gulpfile.js
```diff
@@ -34,88 +34,90 @@ const babel = require('gulp-babel');
...
+	var proc = spawn('./build.sh', [], {cwd: process.cwd()});
+	proc.stdout.on('data', function (data) {
+		process.stdout.write(data);
+	});
+	proc.stderr.on('data', function (data) {
+		process.stderr.write(data);
+	});
+	proc.on('close', function (code) {
+		cb(code)
+	})
```
dstillman: Most/all of the above should just be squashed into https://github.com/zotero/zotero-connectors/pull/51/commits/b4f2aaa0db3fd5a49374bfe248e3565c0970a76c.

build.sh
```diff
@@ -391,6 +384,11 @@ perl -pe "s|<!--SCRIPTS-->|${scripts:8:$((${#scripts}-10))}|s" "$SRCDIR/safari/I
 | perl -000 -p -e 's|(<key>(?:CFBundleShortVersionString\|CFBundleVersion)</key>\s*)<string>[^<]*</string>|$1<string>'"$VERSION"'</string>|sg' \
 > "$BUILD_DIR/safari.safariextension/Info.plist"
 
+# Transpile Safari JS for Safari 10.0<
+echo "Transpiling Safari JS..." >> "$LOG";
+./node_modules/babel-cli/bin/babel.js "$BUILD_DIR/safari.safariextension/" --out-dir "$BUILD_DIR/safari.safariextension/" --presets es2015 -q >> "$LOG" 2>&1
```
dstillman: Here and elsewhere, should be an absolute path using `$CWD` (which is badly named, but we can fix that another time). No guarantee that the script will be run from the root.

gulpfile.js
```diff
@@ -0,0 +1,98 @@
...
+});
+
+gulp.task('watch-chrome', function () {
+	var watcher = gulp.watch(['./src/chrome/**', './src/common/**', './src/safari/**']);
+	watcher.on('change', function(event) {
+		gulp.src(event.path)
+			.pipe(processFile())
+			.on('close', reloadChromeExtensionsTab);
+	});
+});
```
dstillman: Can we just try to do this automatically on macOS, as before, and make it a non-fatal error if chrome-cli fails? A little annoying to have to run a different command.

adomasven: Yeah, but when you develop for safari you have chrome focusing and refreshing extensions page. Not fun.

adomasven: Or firefox, edge, later on.

PR #59, opened 2016-12-15
Chrome/Firefox connector parity with Zotero for Firefox

adomasven:
## Proxying
The connector now supports proxy feature present in Z4FX. A relevant section has been added to the preferences page: 
image

Upon proxy detection standard browser notifications are displayed:
image
Firefox currently does not support notifications with buttons, which means we do not have a button linking to preferences and instead say "go change this in preferences". Ideally we would use the zotero progress window, but as proxy redirection entails loads of redirects and the progress window is embedded on the page, we cannot do that.

We have fully working EZProxy detection, just like in Z4FX. We also have a Juniper proxy detection, which I couldn't test, because I do not have access.

The Safari connector does not support required APIs to do this, however we now have [overall better proxy handling](https://github.com/zotero/zotero/pull/1129) from which Safari benefits too - translation behind proxied resources should be more reliable.

## Importing

The connector now prompts for importing when opening CSL files
image
and displays a progress window if user agrees or prompts for standard download if they decline.
image

It also supports importing RIS/BibTeX files. With these we have a mirroring pref to Z4FX, where import hosts are remembered permanently, once the user accepts an individual host. These are listed in the advanced tab of the preferences:
image

Importing is only supported while Standalone is open.

## Testing on Firefox

- Build like you would for Chrome
- Go to about:debugging#addons
- Click _Load Temporary Add-On_
- Navigate to the build directory and select the `browserExt/manifest.json` file.


src/common/preferences/preferences.html
```diff
@@ -35,53 +35,51 @@
...
-						Zotero Connector has not experienced any errors since it was first started.
-					</div>
-					<div id="general-have-errors" style="display:none">
-						<p>Zotero Connector has experienced the following errors since it was first started:</p>
-						<p><textarea id="general-textarea-errors" style="width:100%; height:10em;" readonly="true"></textarea></p>
-						<p><input type="button" id="general-button-report-errors" value="Report Errors"/></p>
-					</div>
-				</div>
+			<div class="group" style="display: none" id="intercept-and-import">
+				<div class="group-title">File Interception</div>
```
dstillman: This is a little intense. How about "Automatic File Importing"?

src/common/preferences/preferences.html
```diff
@@ -104,6 +102,19 @@
 				</div>
 			</div>
 			<div class="group">
+				<div class="group-title">Report Errors</div>
+				<div class="group-content">
+					<div id="general-no-errors">
+						Zotero Connector has not experienced any errors since it was first started.
```
dstillman: "first" is redundant, and "experienced" is a little anthropomorphic. How about just `No errors have been logged.`?

src/common/preferences/preferences.html
```diff
@@ -104,6 +102,19 @@
...
+					</div>
+					<div id="general-have-errors" style="display:none">
+						<p>Zotero Connector has experienced the following errors since it was first started:</p>
```
dstillman: Maybe `The following errors have occurred since the Zotero Connector was started:`

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,618 @@
...
+	
+	render: function() {
+		return (
+			<div>
+				<label><input type="checkbox" name="transparent" onChange={this.handleCheckboxChange} defaultChecked={this.state.transparent}/>&nbsp;Enable proxy redirection</label><br/>
+				<div style={{marginLeft: "1em"}}>
+					<label><input type="checkbox" disabled={!this.state.transparent} onChange={this.handleCheckboxChange} name="autoRecognize" defaultChecked={this.state.autoRecognize}/>&nbsp;Automatically detect new proxies</label><br/>
+					<label><input type="checkbox" disabled={!this.state.transparent} onChange={this.handleCheckboxChange} name="showRedirectNotification" defaultChecked={this.state.showRedirectNotification}/>&nbsp;Show a notification when redirecting through a proxy</label><br/>
+					<br/>
+					<label><input type="checkbox" disabled={!this.state.transparent} onChange={this.handleCheckboxChange} name="disableByDomain" defaultChecked={this.state.disableByDomain}/>&nbsp;Disable proxy redirection when my domain name contains (only available when Zotero Client is running)</label><br/>
```
dstillman: > Disable proxy redirection when my domain name contains (only available when Zotero Client is running)

I think the part in parens has to be an asterisk below or somehow otherwise separate from this line

Also just "Zotero", not "Zotero Client"

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,618 @@
...
+	},
+	
+	render: function(){
+		var configuredProxies;
+		if (this.state.proxies.length) {
+			configuredProxies = this.state.proxies.map((proxy, i) => 
+				<option value={i} key={i}>{proxy.scheme}</option> 
+			);
+		} else {
+			configuredProxies = <option value={-1}>N/A</option>;
```
dstillman: "N/A" isn't really right here. Can't this just be empty?

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,618 @@
...
+						<label><input type="checkbox" name="dotsToHyphens" onChange={this.handleCheckboxChange} checked={this.state.currentProxy.dotsToHyphens}/>&nbsp;Automatically convert hyphens to dots in proxied hostnames</label><br/>
+						<label style={{visibility: multiHost ? null : 'hidden'}}><input type="checkbox" name="autoAssociate" onChange={this.handleCheckboxChange} checked={this.state.currentProxy.autoAssociate}/>&nbsp;Automatically associate new hosts</label><br/>
+					</p>
+					<p style={{display: "flex", alignItems: "center"}}>
+						<label style={{alignSelf: "center", marginRight: "5px"}}>Scheme: </label>
+						<input style={{flexGrow: "1"}} type="text" name="scheme" onChange={this.handleSchemeChange} value={this.state.currentProxy.scheme}/>
+					</p>
+					<p>
+						You may use the following variables in your proxy scheme:<br/>
+						&#37;h - The hostname of the proxied site (e.g., www.zotero.org)<br/>
```
dstillman: I know this is copied over, but we should probably use "www.example.com" — using zotero.org here is a little misleading.

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,618 @@
...
+		if (this.props.transparent && this.state.currentProxy) {
+			let multiHost = this.state.currentProxy.scheme.indexOf('%h') != -1;
+			let configuredHosts = this.state.currentProxy.hosts.map((host, i) => 
+				<option value={i} key={i}>{host}</option>);
+				
+			proxySettings = (
+				<div className="group" style={{marginTop: "10px"}}>
+					<p style={{display: "flex", alignItems: "center", flexWrap: "wrap"}}>
+						<label><input type="checkbox" name="dotsToHyphens" onChange={this.handleCheckboxChange} checked={this.state.currentProxy.dotsToHyphens}/>&nbsp;Automatically convert hyphens to dots in proxied hostnames</label><br/>
+						<label style={{visibility: multiHost ? null : 'hidden'}}><input type="checkbox" name="autoAssociate" onChange={this.handleCheckboxChange} checked={this.state.currentProxy.autoAssociate}/>&nbsp;Automatically associate new hosts</label><br/>
```
dstillman: "Automatically associate new hosts" should be first, and the hyphens pref should probably be on a new line

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,618 @@
...
+		this.setState({hosts});
+		Zotero.Prefs.set('allowedInterceptHosts', hosts);
+	},
+	
+	render: function() {
+		var hosts;
+		if (this.state.hosts.length) {
+			hosts = this.state.hosts.map((h, i) => <option value={i} key={i}>{h}</option>);
+		} else {
+			hosts = <option value={-1}>N/A</option>;
```
dstillman: Just empty here too

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,618 @@
...
+		if (this.state.currentHostIdx != -1) {
+			hostname = <p style={{display: this.state.currentHostIdx === -1 ? 'none' : 'flex'}}>
+				<label style={{alignSelf: 'center'}}>Hostname: </label>
+				<input style={{flexGrow: '1'}} type="text" defaultValue={this.state.hosts[this.state.currentHostIdx] || ''} onChange={this.handleHostnameChange}/>
+			</p>
+		}
+		
+		return (
+		<div>
+			<p>Available when Zotero is running.</p>
```
dstillman: No period

src/common/utilities.js
```diff
@@ -0,0 +1,44 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+(function() {
+
+"use strict";
+
+Zotero.Utilities = Zotero.Utilities || {};
+
+Zotero.Utilities.debounce = function(fn, delay) {
```
dstillman: Add comment

src/browserExt/mimeTypeHandler.js
```diff
@@ -0,0 +1,181 @@
...
+		} else if (Zotero.Prefs.get('interceptAndImport') && 
+			Zotero.MIMETypeHandler.importMIMETypes.has(contentType)) {
+			
+			let hosts = Zotero.Prefs.get('allowedInterceptHosts');
+			let isEnabledHost = hosts.indexOf(URI.host) != -1;
+			if (isEnabledHost) {
+				Zotero.MIMETypeHandler.importFile(details, 'import');
+			} else {
+				Zotero.MIMETypeHandler.confirm(details, `Would you like to use Zotero to import items from ${URI.host}?\n\n` +
+					'You can manage automatic file import in Zotero Connector preferences.').then(function(response) {
```
adomasven: Yea i'll just get a react component and we will inject over a page. Seems like this is inevitable (although kind of awful to do), given the types of interactions Zotero offers. Popovers not usable for prompts, as they cannot be opened programatically.

src/browserExt/webRequestIntercept.js
```diff
@@ -0,0 +1,128 @@
...
+		beforeSendHeaders: [],
+		headersReceived: [],
+		errorOccurred: []
+	},
+	
+	reqIDToReqHeaders: {},
+
+	init: function() {
+		// Ignore XHR
+		let types = ["main_frame", "sub_frame", "stylesheet", "script", "image", "font", "object", "ping", "other"];
```
adomasven: We do for proxy url rewriting/redirection. Maybe not "ping", although I haven't looked at what that actually is. Although the reason why XHR is ignored escapes me now. Talk about useful comments.

dstillman: Ah, OK. Was just thinking about file interception.

src/common/preferences/preferences.html
adomasven: Yea I've just been moving this around for UI reasons, but fair point. Might want to revise more copy in the connector.

src/common/preferences/preferences.jsx
adomasven: Users manually entering their proxy, which requires dots-to-hyphenization, saving to Zotero and being unable to retrieve attachments OR getting items but the urls end up not unproxied. Depends on what urls the resource provider serves when accessed via proxy. Browsing is unaffected.

Detection automatically sets the pref, so automatically added proxies are good.

Then again if you were willing to write out the scheme for your proxy within the constraints of our kinda arbitrary %-tags, you are likely to know what this pref is about. 

But I can see a casual user wanting to disable automatic host detection for a proxy temporarily, so these two prefs are really on different levels of advancedness.

dstillman: I guess we might as well leave it — most people won't be messing around in this pane anyway. Should definitely switch the order, though.

src/browserExt/mimeTypeHandler.js
```diff
@@ -0,0 +1,181 @@
...
+	 * Send an XHR request to retrieve and import the file into Standalone
+	 */
+	importFile: function(details, type) {
+		chrome.tabs.get(details.tabId, function(tab) {
+			// Make sure scripts injected so we can display the progress window
+			Zotero.Connector_Browser.injectScripts(tab.id).then(function() {
+				Zotero.Messaging.sendMessage('progressWindow.show', type == 'csl' ? 'Installing a style' : null, tab);
+			
+				var xhr = new XMLHttpRequest();
+				xhr.open('GET', details.url);
```
adomasven: Umm.. Let's wait until people are having problems. Getting this to work would need us to cache the POST body of every request, which is likely to cause a whole array of performance related problems. I'll add a comment in the code though.

src/common/preferences/preferences.html
```diff
@@ -35,53 +35,51 @@
...
+				<div class="group-title">Zotero Status</div>
 				<div class="group-content">
-					<p>Zotero Standalone is currently <span id="general-span-standalone-status"></span>.</p>
+					<p>Zotero is currently <span id="general-span-standalone-status"></span>.</p>
 					<p><input type="button" value="Update Status" id="general-button-update-standalone-status"/></p>
 				</div>
 			</div>
 			<div class="group">
-				<div class="group-title">Save via Server</div>
+				<div class="group-title">Save to Zotero.org</div>
```
adomasven: image
image
This is a section title, looks better capitalized imo.

src/browserExt/mimeTypeHandler.js
dstillman: Oh, yeah, that wouldn't be good. Can we determine from here whether the original request was a POST, and not show the dialog if so? There's no reason to think a GET would work for a POST, and if we're retriggering the request on Cancel too (I'm not sure if we are) then this would mean breaking download of such files.

src/common/preferences/preferences.html
dstillman: I don't disagree with you, but there's nothing grosser than capitalized domain names.

Calling it zotero.org also may be a little technical — I've done that a couple places in the client too, but given that people don't use or understand domain names anymore, it may not be the best approach in general. "Save to Web Library" seems a little less clear to me, since the stuff you save to Zotero can end up there too after syncing. Is "Save to Cloud" too buzzwordy? That's sort of the most precise updating of "Save to Server", which I think was clear, if technically inexact.

adomasven: > given that people don't use or understand domain names anymore

They don't? What do people use then?

Judging from the few people that wander into forums every now and then attempting to use Zotero Connectors without realizing there is a client and checking their library on the website, "Save to zotero.org" sounds like the most user friendly name.

For those syncing, who may need to use the connector without the client for whatever reason, "Save to Web Library" may be a little clearer, but "Save to zotero.org" still works. "Save to Server" is confusing. The immediate questions are "What server? Where is said server? Is there a user-friendly way to look at what is saved on this server?"

dstillman: > They don't? What do people use then?

They [definitely do not](https://www.theguardian.com/technology/blog/2010/feb/11/facebook-readwriteweb). They use Google. Which doesn't mean that they wouldn't necessarily understand "Save to Zotero.org", but I do think it's a little technical.

I wasn't suggesting "Save to Server", to be clear — that should definitely be replaced.

adomasven:
Prompt updated to a custom react component:
image

src/common/preferences/preferences.jsx
```diff
@@ -0,0 +1,632 @@
...
+					</p>
+					
+					<div style={{display: "flex", flexDirection: "column", marginTop: "10px"}}>
+						<label>Hostnames</label>
+						<select size="8" multiple onChange={this.handleHostSelectChange} value={[this.state.currentHostIdx]}>
+							{configuredHosts}
+						</select>
+						<p>
+							<input style={{minWidth: "80px", marginRight: "10px"}} type="button" onClick={this.handleHostButtonClick} value="+"/>
+							<input style={{minWidth: "80px", marginRight: "10px"}} type="button" onClick={this.handleHostButtonClick} disabled={this.state.currentHostIdx == -1} value="-"/>
```
dstillman: `Remove` instead of `-`

(The minus alone looks a little weird without a plus.)

src/common/ui/modal-prompt.jsx
```diff
@@ -0,0 +1,206 @@
...
+	
+	getDefaultProps: function() {
+		return {
+			title: "Zotero Connector",
+			checkbox: false,
+			checkboxText: "",
+			checkboxBlock: false,
+			input: false,
+			inputPlaceholder: "",
+			button1Text: "Ok",
```
dstillman: `OK`

src/browserExt/contentTypeHandler.js
```diff
@@ -147,13 +165,13 @@ Zotero.MIMETypeHandler = {
 						Zotero.Connector.callMethod(options, this.response, function(result, status, response) {
 							if (result) {
 								Zotero.Messaging.sendMessage('progressWindow.itemProgress',
-									[Zotero.ItemTypes.getImageSrc('dictionaryEntry'), result.name, null, 100], tab);
+									[chrome.extension.getURL('images/csl-style.png'), result.name, null, 100], tab);
```
adomasven: I'll address this more thoroughly with a separate commit, but I've changed this icon to be 2x and downscaled. Currently the connector icons are a bit of a mess. Some are 2x, some are not, no HiDPI detection at all.

dstillman:
Can you tidy up these commits a bit? E.g., 1f8b98593cb2f2883cfb6c2b945a9007695a5e1f has a lot of unrelated edits to earlier commits. (These can be easily avoided in the future with [auto-squash/fixup](https://robots.thoughtbot.com/autosquashing-git-commits). Then they're still recorded as new commits but they can all be automatically squashed with a single `rebase` command.)

adomasven:
>(These can be easily avoided in the future with auto-squash/fixup. Then they're still recorded and pushed as new commits but they can all be automatically squashed before merging with a single rebase command.)

Thanks for the heads-up for this one. Yea, sure. Otherwise looks fine?

dstillman:
Yup, looks great.

PR #69, opened 2017-01-04
Feature/prompt on first click

adomasven:
- Remove the previous behaviour of prompting on install on firefox
- If upon install if Zotero client is not running, local repo has not successfully been checked before and user has not interacted with the extension yet - display the Z logo
- On click of Z logo open tab either to https://www.zotero.org/start_connector or https://www.zotero.org/support/zotero_for_firefox_migration depending on browser.


src/browserExt/background.js
```diff
@@ -366,8 +385,16 @@ Zotero.Connector_Browser = new function() {
...
 	
-	function _save(tab) {
-		if(_tabInfo[tab.id] && _tabInfo[tab.id].translators && _tabInfo[tab.id].translators.length) {
+	function _browserAction(tab) {
+		if (Zotero.Prefs.get('firstUseNoClient')) {
+			Zotero.Prefs.set('firstUseNoClient', false);
+			if (Zotero.isFirefox && Zotero.Prefs.get('previousConnectorVersion') < '5') {
+				Zotero.Connector_Browser.openTab('https://www.zotero.org/support/zotero_for_firefox_migration');
+			} else {
+				Zotero.Connector_Browser.openTab('https://www.zotero.org/start_connector');
```
dstillman: Can use `/connector_start` for everything. We don't want to show new Firefox users a page that doesn't apply to them. We can detect Firefox on that page and show a section at the top with a link to another page.

src/browserExt/background.js
```diff
@@ -235,6 +243,18 @@ Zotero.Connector_Browser = new function() {
 		}
 	}
 	
+	function _showFirstTimeUI(tab) {
+		chrome.browserAction.setIcon({
+			tabId: tab.id,
+			path: "images/zotero-new-z-16px.png"
```
dstillman: Should use a higher-res icon here, even if it's scaled on low-res displays.

dstillman:
Are you sure this works? `connector.repo.lastCheck.localTime` is just the local time from the last repo check, whether that was from z.org or the client, so it would get set as soon as the connector first pings z.org. So not sure anyone would ever see the icon. I don't see it, at least, if I delete all the localStorage prefs manually and then reload the extension.

src/browserExt/background.js
adomasven: Do we have a 32px version of this without the background white corner thing?

adomasven:
>Are you sure this works? I don't see it, at least, if I delete all the localStorage prefs manually and then reload the extension.

Yeah this is bugged, although deleting all localStorage prefs should have worked. Note that localStorage is URL namespaced, so you would actually have to delete localStorage prefs on the background page debugger for the extension for this to work.

I incorrectly assumed that `connector.repo.lastCheck.remoteTime` is the value used for updates from repo, and `connector.repo.lastCheck.localTime` is the value used for client updates.

dstillman:
Yes, I deleted from background page. As soon as you load any page it loads translators from the repo and sets that pref.

src/browserExt/background.js
dstillman: Oh, no, I guess we don't. Oh well. We can replace once we make a new icon.

PR #114, opened 2017-04-19
Fix #111

### LINKED ISSUE #111: If proxy redirects to unproxied site, Zotero should not attempt to redirect again
When I'm on campus, attempting to access a site through the MIT proxy will result in the proxy redirecting me back to the original site. After the proxy redirects me to the original site, the Firefox connector is trying to send me back through the proxy, e.g.:

```
connector => true, version => 4.999.17, platform => MacIntel, locale => en-US, userAgent => Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:52.0) Gecko/20100101 Firefox/52.0, appName => Firefox Connector, zoteroAvailable => true

=========================================================

(3)(+0005439): Proxies.properToProxy: http://science.sciencemag.org/content/353/6301/757.full to http://science.sciencemag.org.libproxy.mit.edu/content/353/6301/757.full

(3)(+0000002): NOTIFICATION: science.sciencemag.org was automatically redirected through science.sciencemag.org.libproxy.mit.edu

(3)(+0000911): Proxies.properToProxy: http://science.sciencemag.org/content/353/6301/757.full to http://science.sciencemag.org.libproxy.mit.edu/content/353/6301/757.full

(3)(+0000001): NOTIFICATION: science.sciencemag.org was automatically redirected through science.sciencemag.org.libproxy.mit.edu

(3)(+0001102): Proxies.properToProxy: http://science.sciencemag.org/content/353/6301/757.full to http://science.sciencemag.org.libproxy.mit.edu/content/353/6301/757.full

(3)(+0000001): NOTIFICATION: science.sciencemag.org was automatically redirected through science.sciencemag.org.libproxy.mit.edu

...
```

Eventually, Firefox gives up and shows me a "The page isn’t redirecting properly" message, and I can't get to the page without turning off proxy redirection. There is [code](https://github.com/zotero/zotero-connectors/blob/b87084cd5cccd7527b0023469685e34f25056e13/src/common/proxy.js#L211-L224) to handle this case, and it works fine as implemented in classic Zotero for Firefox, but it doesn't seem to be working in the WebExtension.

simonster:
- Keep track of request IDs that have already been redirected, to avoid redirecting multiple times
- Separate proxy-related listeners into one that observes the request (to determine whether it should be redirected) and one that observes the response (to determine whether to add new proxies), so that we don't need to save request headers for every request

### DIFF
--- a/src/browserExt/webRequestIntercept.js
+++ b/src/browserExt/webRequestIntercept.js
@@ -38,26 +38,25 @@ Zotero.WebRequestIntercept = {
 		errorOccurred: []
 	},
 	
-	reqIDToReqHeaders: {},
+	reqIDToReqMeta: {},
 
 	init: function() {
 		// Ignore XHR because we use those to fetch intercepted resources
 		let types = ["main_frame", "sub_frame", "stylesheet", "script", "image", "font", "object", "ping", "other"];
 		chrome.webRequest.onBeforeSendHeaders.addListener(Zotero.WebRequestIntercept.handleRequest('beforeSendHeaders'), {urls: ['<all_urls>'], types}, ['blocking', 'requestHeaders']);
-		chrome.webRequest.onErrorOccurred.addListener(Zotero.WebRequestIntercept.removeRequestHeaders, {urls: ['<all_urls>'], types});
-		chrome.webRequest.onCompleted.addListener(Zotero.WebRequestIntercept.removeRequestHeaders, {urls: ['<all_urls>'], types});
+		chrome.webRequest.onErrorOccurred.addListener(Zotero.WebRequestIntercept.removeRequestMeta, {urls: ['<all_urls>'], types});
+		chrome.webRequest.onCompleted.addListener(Zotero.WebRequestIntercept.removeRequestMeta, {urls: ['<all_urls>'], types});
 		chrome.webRequest.onHeadersReceived.addListener(Zotero.WebRequestIntercept.handleRequest('headersReceived'), {urls: ['<all_urls>'], types}, ['blocking', 'responseHeaders']);
-		
+
 		Zotero.WebRequestIntercept.addListener('beforeSendHeaders', Zotero.WebRequestIntercept.storeRequestHeaders)
 	},
+
+	storeRequestHeaders: function(details, meta) {
+		meta.requestHeadersObject = details.requestHeadersObject;
+    },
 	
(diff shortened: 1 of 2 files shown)


simonster:
(FWIW, the issue was that `originUrl` is not actually defined for redirects, only for resource loads. According to the [MDN WebExtension docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onHeadersReceived), `originUrl` is also not supported in Chrome.)

adomasven:
- There is at least [one point](https://github.com/zotero/zotero-connectors/blob/21f400eaba4c987047ca0d507b923d55cf8cc99b/src/common/proxy.js#L703) in the proxy detector where request headers are needed. That was `responseHeadersObject['referer']`, but it was a bug, I've just committed a fix. Technically we could just add `referer` to `meta` now, but I've been storing the headers in anticipation of other features, including fx-parity, needing those, so this is a bit of a step back.

- What's wrong with [this](https://github.com/zotero/zotero-connectors/commit/c03d7573063f8d938255f18f920e4a484ea308f3#diff-6d0ffc82769922b40a8a73f30989a879R249) solution?

simonster:
- If there is a legitimate reason, I guess we can keep the response headers around. In terms of perf, the profiler indicates it is unlikely to make a noticeable difference. I just hope the browser really does fire the events we listen to in order to clean things up. (We might want to be more mindful about not calling `url.parse` unless a proxy actually matches, though, since we are spending ~45 ms there when loading Netflix, and it doesn't seem necessary to call it 3x on `details.url` for every request.)
- While redirecting from `onBeforeSendHeaders` works in Firefox, it isn't actually allowed by the WebExtensions API, which only allows redirects from `onBeforeRequest` and `onHeadersReceived`, and doesn't work in Chrome. I should probably add a comment to this effect.

simonster:
Okay, here's a smaller version of the fix. I noticed that we still don't really need both the request and response headers simultaneously to handle EZProxy detection (the cases where we use the response headers are still independent of the cases where we use the request headers) but if there's some planned functionality that will require both, I don't want to interfere.

I made a couple of other tiny tweaks to reduce the performance impact of the proxy code. There is still more that could be done, but the overhead is unlikely to be meaningful if we can stop observing script, image, font, and stylesheet loads (#115).

PR #117, opened 2017-04-25
Custom notifications for proxies

### LINKED ISSUE #112: We should not add new proxies without confirmation by the user
[The code](https://github.com/zotero/zotero-connectors/blob/b87084cd5cccd7527b0023469685e34f25056e13/src/common/proxy.js#L174-L175) states that the notifications API only allows this on Chrome, but it's really critical that we ask for user confirmation before adding new proxies, even if it's not through the notifications API. Otherwise it's possible for a malicious site to add a new proxy such that mybank.com redirects to them. We don't necessarily need confirmation to add sites to existing proxies, since presumably they are trusted.

### LINKED ISSUE #113: Proxy notifications are annoying
It's a little irritating that proxy redirect notifications make the computer beep, and that there's not enough space in the window to show all of the relevant information. It would be nice if we could do an overlay on the page instead of using the notifications API, although this may be more challenging to implement.

adomasven:
Closes #112, closes #113 

image

UI changes/suggestions welcome

### DIFF
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -216,6 +216,10 @@ Zotero.Connector_Browser = new function() {
 		}
 	};
 	
+	this.openPreferences = function(paneID, tab) {
+		this.openTab(chrome.extension.getURL(`preferences/preferences.html#${paneID}`), tab);
+	};
+	
 	this.notify = function(text, buttons, timeout=15000, tab=null) {
 		// Get current tab if not provided
 		if (!tab) {
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -237,7 +237,7 @@ Zotero.Connector_Browser = new function() {
 					// If it still exists try again
 					// But make sure translation scripts are injected first
 					return this.injectTranslationScripts(tab)
-						.then(resolve(this.notify(text, buttons, timeout, tab)));
+						.then(() => resolve(this.notify(text, buttons, timeout, tab)));
 				}.bind(this));
 			}.bind(this))
 		}.bind(this));
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -160,14 +160,23 @@ Zotero.Connector_Browser = new function() {
 	 * @returns {Promise} A promise that resolves when all scripts have been injected
 	 */
 	this.injectTranslationScripts = function(tab, frameId=0) {
-		let deferred = Zotero.Promise.defer();
+		// Prevent triggering multiple times
+		let key = tab.id+'-'+frameId;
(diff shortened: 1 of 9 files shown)


dstillman:
I'll make some UI changes in a follow-up commit (though the text can be left-justified to start).

This isn't showing for me at all in Firefox at the moment, so I'll just ask, what happens when Proxy Settings is clicked? Does the notification stay open, so that they can still make the choice, or is it equivalent to Dismiss? (The former seems better.)

I think Accept should probably be the right-most option, since generally speaking people are going to want to accept this. And maybe we want to say "Ignore" instead of "Dismiss"? If they ignore, will they get this again the next time they authenticate through the proxy for this site?

(I understand that some of these questions may have applied to the previous implementation too.)

src/browserExt/background.js
```diff
@@ -206,6 +215,29 @@ Zotero.Connector_Browser = new function() {
...
+		return Zotero.Messaging.sendMessage('notify', [text, buttons, timeout, tab.status], tab).then(function(response) {
+			if (response != undefined) return response;
+			// Tab url changed or tab got removed, hence the undefined response
+			return new Zotero.Promise(function(resolve) {
+				chrome.tabs.get(tab.id, function(tab) {
+					if (!tab) return;
+					// If it still exists try again
+					// But make sure translation scripts are injected first
+					return this.injectTranslationScripts(tab)
+						.then(resolve(this.notify(text, buttons, timeout, tab)));
```
adomasven: Good spot!

adomasven:
> This isn't showing for me at all in Firefox at the moment, so I'll just ask, what happens when Proxy Settings is clicked? Does the notification stay open, so that they can still make the choice, or is it equivalent to Dismiss? (The former seems better.)

The notification is dismissed. Not ideal, but these at the moment have a timeout, so I figured they'd disappear anyway. And I think we should really have a timeout. Also the fact that it is dismissed is less relevant for other notifications, where it just tells you that you have been redirected.

> If they ignore, will they get this again the next time they authenticate through the proxy for this site?

Yes. Hence the Proxy Settings button, so they can disable this if not needed. That's both the previous and probably the best solution.

dstillman:
Oh, I don't think this one should have a timeout. This is not an easy notification to parse, and particularly if we're asking for an answer from the user, having it go away before they've answered doesn't seem good. For the non-interactive ones, a timeout makes sense (and 10 seconds might actually be too long).

adomasven:
Bump

adomasven:
image

dstillman:
Merged in 7f5bec59cb8d07b2f8c059c32a5e9ca6ecbfb5bd

PR #129, opened 2017-05-16
Prompt with additional details for new proxy

### LINKED ISSUE #128: Restore prompt with additional details before adding new proxy
This is what we showed in Zotero 4.0 after you clicked "Enable…" on the proxy notification, with a time delay on the "Add Proxy" button:

warning

In the 5.0 connectors we just have an Accept button on the notification, with no further prompt.

It's possible the prompt is overkill, since users do already have to go out of their way to accept the proxy, but people are fairly inclined to click Accept to make annoying things go away. Of course, if they're inclined to do that, maybe they'd click through the prompt too, but given that this is a fairly dangerous feature, the additional "Only add proxies linked from your library, school, or corporate website" seems worthwhile.

adomasven:
Closes #128 

Z4.0 prompt:
warning

New prompt:
warning

I think we need Zotero branding on the prompts.

### DIFF
--- a/src/common/messaging.js
+++ b/src/common/messaging.js
@@ -113,6 +113,11 @@ Zotero.Messaging = new function() {
 		}
 		// Use the promise or response callback in BrowserExt for advanced functionality
 		else if(Zotero.isBrowserExt) {
+			// Get current tab if not provided
+			if (!tab) return new Zotero.Promise(function(resolve) {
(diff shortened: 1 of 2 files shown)


dstillman:
> I think we need Zotero branding on the prompts.

Agreed. Let me see what I can do.

Issue #93, opened 2017-01-24
Use APIv3 for all requests


dstillman:

adomasven:
Closed with https://github.com/zotero/zotero-connectors/pull/143

Issue #135, opened 2017-05-23
Integration tests


adomasven:
This is becoming pretty critical now. We can get away without unit tests, but we must ensure basic functionality doesn't accidentally break.

Platforms:
- [ ] Chrome (Latest)
- [ ] Firefox (Latest)
- [ ] Firefox (ESR)
- [ ] Safari (Latest)
- [ ] Safari (Oldest-supported)

Having at least Firefox ESR would be a good start.

Functionality:
- [ ] Debug output submission
- [ ] Report ID submission
- [ ] Translation to Zotero
- [ ] Translation to Server
- [ ] Prompts
- [ ] Notifications
- [ ] Proxy detection
- [ ] Style installation
- [ ] Data-files import

adomasven:
So before I begin, this is the plan:
- Use the [node.js Selenium Webdriver](http://seleniumhq.github.io/selenium/docs/api/javascript/index.html) to drive browsers
- Use the [Mocha](https://mochajs.org/)-[Chai](http://chaijs.com/) testing framework (currently used by the Zotero client unit-test suite)
- Use [Express.js](https://expressjs.com/) to mock responses from the client, zotero.org and translation targets.

I'll start with Firefox ESR. Any objections/suggestions @dstillman @tnajdek @inukshuk

adomasven:
The plan has changed to containing the tests within the extension with the debugging build. The webdriver spec and implementations may be ready in 5 years to build something reasonable, but not today. We might still use Selenium to actually launch the browsers with extensions preloaded, but that's not strictly necessary. Express.js had been ditched too as we can stub responses with sinon.js since the tests run in the inside the blackbox.

adomasven:
Safari automation is stuck [upstream](https://github.com/travis-ci/travis-ci/issues/7187). High Sierra (release in 2 weeks) required to launch a session with automation enabled and then it will probably take a while for travis-ci to support it.

Either way, I'll work on tests and we can at least run them manually before releases, which will still catch bugs.

adomasven:
Seems like we might be able to set up travis automation now, see upstream link above.

adomasven:
We should consider rewriting these since they rely on eval that is no longer viable with MV3, but there have been improvements to headless browsers and selenium since writing this that would allow to automate the browser and get its contents directly in nodejs.

mbrukman:
[...]

Issue #123, opened 2017-05-08
Add Zotero.HTTP.request() and deprecate doGet and doPost


adomasven:
@dstillman in #122 :
> (Only semi-related, and we can stuff support for this into the existing function in the interim, but we should probably just port Zotero.HTTP.request() from the client over to the connectors, relying on the native Promise in the bookmarklet, and deprecate doGet/doPost in translators. Of course, it would be nice to use async/await with that, but that might need to wait a bit longer.)

dstillman:
Done in https://github.com/zotero/zotero-connectors/commit/d52dab1b2932332bcb1faedbd633d775c38cbc6e

PR #143, opened 2017-07-11
Integration tests for the connector on Travis-CI

adomasven:
Addresses #135 

image

In this PR:
- Tests for Preferences and Translation
- Refactoring some older bits of the codebase, including all API requests that were using v2 of the API.
- [Travis-CI](https://travis-ci.org/adomasven/zotero-connectors) testing with Chrome, Firefox and Firefox-ESR

Notes:
- Testing for Safari are a bit complicated, as current tests use a lot of `chrome.` API calls. Maybe later.
- There are more things to test, but since the framework is in place additional tests can always be added later, as required.


adomasven:
Updated to close #144

adomasven:
Closes #93

adomasven:
Progress towards #123

src/common/test/data/journalArticle-single.html
```diff
@@ -0,0 +1,53 @@
...
+<script src='/inject/translate_inject.js'></script>
+<script src='/messages.js'></script>
+<script src='/messaging_inject.js'></script>
+<script src='/api.js'></script>
+<script src='/inject/inject.js'></script>
+<script src='/test/testSetup.js'></script>
+<script src='/lib/sinon.js'></script>
+<script src='/lib/react.js'></script>
+<script src='/lib/react-dom.js'></script>
+<script src='/ui/modal-prompt.js'></script>
```
adomasven: https://github.com/zotero/zotero-connectors/pull/143/files/57024974d0e5fc4e9b8ae657a8246724f3fda683..3665c6e54232ae4d183beb51ecc19a3059d21d32#diff-9e0e8bc9053bf2e3214ad5175f823c6eR12

We could maybe do this in the build step, but some of this is "programatically injected" during normal operation. I don't have strong feelings.

src/common/http.js
```diff
@@ -28,175 +28,176 @@
...
+	* @param {N/A}				cookieSandbox	Not used in Connector
+	* @param {Object}			headers			HTTP headers to include with the request
+	* @return {Boolean} True if the request was sent, or false if the browser is offline
+	*/
+	this.doGet = function(url, onDone, responseCharset, cookieSandbox, headers) {
+		Zotero.debug('Zotero.HTTP.doGet is deprecated. Use Zotero.HTTP.request');
+		this.request('GET', url, {responseCharset, headers})
+		.then(onDone)
+		.catch(function(e) {
+			onDone({status: e.status, responseText: e.responseText});
```
dstillman: I think technically this should be called in the second argument to the previous `then`. Here it would be called on a non-HTTP error if `onDone(xmlhttp)` threw an error in the `then`.

adomasven:
Finally got this one to work again properly. Changes implemented

PR #148, opened 2017-07-18
Show an error when collection not editable

### LINKED ISSUE #95: Saving while feed selected works and saves to library.
We should either correctly indicated that the item is being saved into "My Library" or prevent saving. The bug might be present for other configuration, where saving falls back to library, while showing a different selected collection.

adomasven:
Closes #95 

Corresponding changes for zotero/zotero available at https://github.com/adomasven/zotero/commit/98cdea0666abc818077998bedd8fff3cf272f3d1


dstillman:
These are both fine with me.

Issue #86, opened 2017-01-12
Add "Save Link to Zotero" context menu


dstillman:
In Firefox this saved either a webpage item or a top-level file (e.g., PDF). For the former, it'd be nice to do automatic translation falling back to webpage (like we do for feed items). The latter is more important, since it'd be a replacement for the Firefox save-dialog modification for people not viewing PDFs in the browser.

adomasven:
We might want to restrict where we show other context-menu options for the connectors, and only show this option on link right-clicks. That way the option will not be hidden in a Zotero Connector submenu.
BrowserExt: https://developer.chrome.com/extensions/contextMenus#type-ContextType
Safari: https://developer.apple.com/library/content/documentation/Tools/Conceptual/SafariExtensionGuide/AddingContextualMenuItems/AddingContextualMenuItems.html

adomasven:
What is the expected behaviour for pages where the primary translator returns "multiple" and we would show an item selection window?

adomasven:
Okay, so after playing around with this a bit more I found out that we are very limited by CSP here. The "hiddenBrowser" feature for `Zotero.HTTP.processDocument` is implemented as an iframe on the injected page and content outside of the current domain is loaded via XHR and then manually written into the iframe. 

The main implication is that `document.location.href` attribute will be set to the parent page. I.e. if you do 'Save Link to Zotero' on scholar.google.com and the resource is at jstor.com, the `document.location.href` attribute will be the scholar.google.com one. At the moment there are 224 mentions of `document.location.href` in various translators, mostly to add a snapshot attachment. Thus items saved in such way fetch wrong snapshots.

We could still offer this for same-origin links or only offer the feature when Zotero is running and do the translation within Zotero.

zuphilip:
We could also try to get rid of most `document.location.href` in translator codes, because they are normally not needed and one can substitute this with the `url` parameter of the `detectWeb` and `doWeb` functions. Would this be helpful?

adomasven:
I think translators should be able to rely on all properties of `document` being correct as it is the only interface used in translation to access the page contents. We should figure this out within Zotero.

adomasven:
Also a relevant issues here: zotero/zotero#348 and #10. Simon had an idea for switching processDocuments to XHR on Zotero, although with XPCOM you could probably work around the incorrect `document.location` property. Not sure if he had any better ideas for implementing it for the connectors.

simonster:
[...]

PR #159, opened 2017-08-03
SSE client

### LINKED ISSUE #102: Context-menu option to add current page as attachment
This would restore functionality of the "Attach Snapshot of Current Page" and "Attach Link to Current Page" options in Zotero for Firefox. (Dragging the link or icon works, but it's annoying if you don't use Cmd-Tab.)

https://forums.zotero.org/discussion/comment/269248/#Comment_269248

adomasven:
Adds SSE support and the first feature we're using it for - attaching current page as a snapshot or link to a selected item.

Closes #102

### DIFF
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -55,7 +55,7 @@ Zotero.Connector_Browser = new function() {
 		var isPDF = contentType == 'application/pdf';
 		_tabInfo[tab.id] = {translators, instanceID, isPDF};
 		
-		_updateExtensionUI(tab);
+		this._updateExtensionUI(tab);
 	}
 	
 	/**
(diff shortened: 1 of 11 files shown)


adomasven:
Requires a modification to reports endpoint to accept plaintext, because this update significantly improves the look of error and debug reports and also includes error/debug info from the client if it is running. The text of the report is generated on the client side.

See https://github.com/zotero/zotero-connectors/pull/159/commits/000978c5e1c2df6c58fccf32df190eb7b4b72355#diff-f4d3a666fd0dcb74381901a1aeb1044eR131

adomasven:
Also, the zotero/zotero PR needs to be merged first, and then this PR needs to update the submodule and a zotero src file reference in the gulpfile.

adomasven:
Updated

adomasven:
Closes #86

adomasven:
Closes #163

PR #178, opened 2017-09-12
Save options submenu and "Reload via Proxy"

dstillman:


dstillman:
Save menu:

save-menu

Proxy menu:

proxy-menu

adomasven:
> We probably want to do this in Safari too, though I'm not sure if
there's an item limit there.

Safari's menu is not limited but currently there is no code to easily add submenus. We might want it later down the line for consistency's sake, although there's not as many options in that connector.

Issue #155, opened 2017-07-28
Show PDF icon for page with PDF iframe and no translator


dstillman:
https://github.com/zotero/zotero-connectors/issues/152#issuecomment-318569335

dstillman:
(If there are multiple we can save the first one, but that will almost never happen.)

dstillman:
The downside, I guess, is that we would have to run a `document.querySelector('iframe')` call on every page without a translator, but hopefully that would be very fast.

dstillman:
The implementation in c3793436875830d083794438941105daeb39aa50 seems way more complicated than I was thinking — I was assuming we would just look for an iframe with a URL with a path (not counting query string) that ended in `.pdf` and hope for the best.

But it also doesn't seem to be working for me. E.g., I'm not seeing a PDF icon on either https://www.cs.tut.fi/~jkorpela/html/iframe-pdf.html or https://pdfobject.com/static.html.

The latter page also shows a couple other techniques (`object` and `embed`) that can be used for PDF embedding, both using `type="application/pdf"`, that I imagine we could easily look for with `querySelector` from the main page.

adomasven:
Both pages work for me on Chrome, Firefox and Safari.

dstillman:
Working for me now. I guess maybe I was testing with Zotero closed? To support that I guess we'd have start sending URLs and cookies up to the API (and hoping sites don't have same-IP restrictions on cookies)?

adomasven:
Detection works fine with Zotero closed. We don't support saving standalone PDFs to server now regardless of whether they are in a frame or not.

dstillman:
PDF detection does not work for me in either Chrome or Firefox with Zotero closed. (It works if I just close Zotero while it's already showing a PDF icon and reload, but not if I close the tab and reopen it.)

dstillman:
Well, now it is, with the latest commits. It wasn't previously.

adomasven:
> Detection works fine with Zotero closed.

Ok, so it actually doesn't, but it doesn't really matter, because PDF saving to zotero.org doesn't either. We will revisit this once PDF saving to zotero.org is available.

Issue #152, opened 2017-07-24
Add "Save to Zotero (PDF)" to Zotero Connector submenu when right-clicking on a framed PDF


dstillman:
This is useful when viewing a framed PDF that isn't recognized by a translator, since it avoids having to do This Frame -> Show Only This Frame in Firefox (and that's not even an option in Chrome?).

dstillman:
This is `fx-parity` due to the removal of the open/save dialog modification. Previously, you could just click the download icon in the PDF viewer and save to Zotero that way.

adomasven:
Do we have any example URLs?

dstillman:
https://www.cs.tut.fi/~jkorpela/html/iframe-pdf.html

adomasven:
Seems like we need to talk upstream if we want this. Chrome does not show addon context-menus in PDFs completely. Firefox does, but the finest way you can filter these are either by showing to all [frames](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/contextMenus/ContextType) or matching [url regex](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/contextMenus/create).

Although if we're taking this upstream we may as well just ask for #83

dstillman:
<blockquote>Chrome does not show addon context-menus in PDFs completely</blockquote>
I have a 1Password menu item when right-clicking on a framed PDF in Chrome (but probably for all frames).

> Although if we're taking this upstream we may as well just ask for #83
I already did — they're very unlikely to do something tied to the current UI in that way. But there might be other options involving download notifications (which may or may not work for us).

Can we tell whether there's a PDF in an iframe from the parent page? What about adding a "Save to Zotero (PDF)" option to our own context menu in that case?

adomasven:
> I have a 1Password menu item when right-clicking on a framed PDF in Chrome (but probably for all frames).

Oh yeah, that actually works.

> Can we tell whether there's a PDF in an iframe from the parent page? What about adding a "Save to Zotero (PDF)" option to our own context menu in that case?

We could check for frames with content-type `application/pdf` with web request intercepting. We could then add a "Save to Zotero (PDF)" with context set to frames, but that would add that context-menu option to all frames, including non-pdf ones. (Seems like we could actually do this properly in [safari](https://developer.apple.com/library/content/documentation/Tools/Conceptual/SafariExtensionGuide/AddingContextualMenuItems/AddingContextualMenuItems.html))

[...]

Issue #56, opened 2016-12-14
Respect automatic tagging option when saving via server


adomasven:
https://forums.zotero.org/discussion/comment/266420#Comment_266420

PR #193, opened 2017-10-10
Generic EM translation in Zotero

adomasven:
Blocked by https://github.com/zotero/translators/pull/1434

In this PR:
1. Make EM the generic translator as described in [previous discussion](https://github.com/zotero/translators/issues/1092#issuecomment-333475693)
2. Make the connector pref for saving snapshots and attachments apply to both saving to Zotero and zotero.org
3. Expose the snapshots pref in the context menu
4. Add submenu and checkbox support to safari popover

Note:
No migration of prefs from the client, since it is not clear whether the saving to zotero.org or saving to Zotero prefs should take precedence, but we have the zotero.org prefs already available.

image

Should make sure that we no longer apply the snapshot and attachment pref in the client.


adomasven:
Effectively reverts https://github.com/zotero/zotero-connectors/issues/155 (Show PDF icon for page with PDF iframe and no translator). Also see the discussion https://github.com/zotero/zotero-connectors/issues/152

adomasven:
Closes #206

rmzelle:
(so I'd like to see, in addition to "Expose the snapshots pref in the context menu" with the "Include Snapshots" toggle, a similar toggle for "Include PDFs and Other Files")

adomasven:
Probably makes sense to move the pref for automatic tags to the connector too https://github.com/zotero/zotero-connectors/issues/56

PR #203, opened 2017-11-15
fix segment fault during convert

petronny:
On Archlinux with gcc 7.2.0-3 and imagemagick 6.9.9.22-1
I get segment fault when converting `src/zotero/chrome/skin/default/zotero/treeitem@2x.png` to `build/safari.safariextension/images/toolbar/treeitem@2x.png`
I removed the `-flatten` flags and it works now. The converted icon is
treeitem 2x
(PS. it's almost transparent)

Is the converted icon correct?

And I created this PR to try to convert with the `-flatten` flags then without it.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -212,8 +212,12 @@ if [ $RETVAL == 0 ]; then
 	cp $IMAGES $PREFS_IMAGES "$BUILD_DIR/safari.safariextension/images"
 	for f in $ICONS
 	do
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks, but a segfault for a standard flag would suggest a bug somewhere in Arch, so that wouldn't be a reason we would change this. I'm not familiar with the different flags, but this is working and doing what we want on our Amazon Linux (~CentOS) build systems.

PR #227, opened 2018-03-30
Google Docs Integration

adomasven:
Adds the submodule for google docs integration.

`integration/connectorIntegration.js` enables anyone to write their own
Zotero integration plugins for web-based editors by exposing the
Integration API to the webpage.  Some security considerations are in order.

TODO: 
- Safari support.
- Move the Google Apps Script API back-end to the Zotero account (currently hosted on mine)
- Testing with multiple doc contributors
- Connector preference to disable the plugin (and disable by default and mark as experimental)
- Make https://www.zotero.org/google-docs lead somewhere

### DIFF
--- a/src/common/integration/connectorIntegration.js
+++ b/src/common/integration/connectorIntegration.js
@@ -1,0 +1,102 @@
+/*
+	***** BEGIN LICENSE BLOCK *****
+	
+	Copyright © 2017 Center for History and New Media
+					George Mason University, Fairfax, Virginia, USA
+					http://zotero.org
+	
+	This file is part of Zotero.
+	
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+	
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+	
+	***** END LICENSE BLOCK *****
+*/
+
+/**
+ * An integration interface intended to work via the Zotero client.
+ * See server_connectorIntegration.js in zotero codebase for expected behaviour
+ */
+Zotero.ConnectorIntegration = {
+	init: function() {
+		// Limit to google docs for now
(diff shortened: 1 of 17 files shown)


src/common/integration/connectorIntegration.js
```diff
@@ -0,0 +1,100 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+/**
+ * An integration interface intended to work via the Zotero client.
+ * See server_connectorIntegration.js in zotero codebase for expected behaviour
+ */
+Zotero.ConnectorIntegration = {
+	init: function() {
+		window.addEventListener('Zotero.Integration.execCommand', async function(e) {
```
dstillman: `e` -> `event`, to distinguish from our use of `e` for errors

Unless there's a technical reason why we can't, we should use `window.postMessage()` and [restrict `event.origin`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Security_concerns) to `docs.google.com` for now. We can consider relaxing this later, maybe by hard-coding sites on request, but we shouldn't let random websites trigger actions in Zotero, even if it's just opening the integration dialog.

adomasven: 1. We [already use](https://github.com/adomasven/zotero-connectors/blob/48c89eee0f545a288f4d0e34984fabd6067f70c5/src/common/inject/inject.jsx#L193) window events to talk to external pages.
1. Events, as opposed to messages, mean that you do not have to handle other messages that pages might post to iframes or other extensions. With a custom event name we only need to handle communications intended for Zotero.

We can restrict the origin by not attaching these event listeners on domains other than `docs.google.com` for now.

dstillman: `ZoteroItemUpdated` doesn't trigger any user-visible behavior though, so that's not really comparable.

Only attaching to `docs.google.com` for now sounds good.

adomasven:
Updated

src/common/integration/connectorIntegration.js
```diff
@@ -0,0 +1,100 @@
...
+		} catch (e) {
+			// Usual response for a request in progress
+			if (e.status == 503) {
+				Zotero.debug(e.message);
+				return;
+			} else if (e.status == 0) {
+				Zotero.Inject.confirm({
+					title: "Is Zotero Running?",
+						message: `
+							The Zotero Connector was unable to communicate with the Zotero desktop application. The Zotero desktop application is required to cite with ${Zotero.clientName}.
```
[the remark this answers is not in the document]
gracile-fr: (I am translating this now and "web-based citing" is very elliptical in my opinion. I'd prefer "Google Docs" or whatever relevant in the context, as you wrote.)

PR #275, opened 2019-02-08
Use the browserExt API to provide the shortcut for GDocs citing

adomasven:
See https://github.com/zotero/zotero-google-docs-integration/issues/24

The benefit of using the browserExt API is that the browser hopefully will not allow invalid shortcuts. The drawbacks is that you cannot use both Ctrl and Alt in a shortcut (which is what our default one does), and if you override a default shortcut (e.g. Ctrl+Shift+C, which is also a shortcut for "Select an element on the page to inspect") then it will not work on any page, not just Google Docs.

Our other option to implement this would be to use our own custom prefs, but that would make it harder to find and more error prone, since we'd have to expose a JSON object to configure it.

Also for some reason the default shortcut (Ctrl+Shift+C) did not automatically get assigned for me on Chrome, perhaps because it overrides a Chrome shortcut, but in that case it falls back to the original shortcut (Ctrl+Alt+C).

image

### DIFF
--- a/src/browserExt/manifest.json
+++ b/src/browserExt/manifest.json
@@ -54,6 +54,13 @@
 			"suggested_key": {
 				"default": "Ctrl+Shift+S"
 			}
+		},
+		"cite": {
+			"suggested_key": {
+				"default": "Ctrl+Shift+C",
+				"mac": "Command+Ctrl+C"
+			},
+			"description": "Add/Edit Citation on Google Docs"
 		}
 	},
 	"minimum_chrome_version": "55",
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -424,6 +424,11 @@ Zotero.Connector_Browser = new function() {
 		return this.notify(text, buttons, seenTimeout, tab);
 	};
 
+	this.getShortcuts = function() {
+		Zotero.debug('getting shortcuts');
+		return browser.commands.getAll();
+	};
+
 	/**
 	 * Update status and tooltip of Zotero button
 	 */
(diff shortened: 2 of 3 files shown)


src/browserExt/manifest.json
```diff
@@ -54,6 +54,13 @@
 			"suggested_key": {
 				"default": "Ctrl+Shift+S"
 			}
+		},
+		"cite": {
+			"suggested_key": {
+				"default": "Ctrl+Shift+C",
+				"mac": "Command+Ctrl+C"
```
dstillman: From [the `commands` documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/commands):

> On Macs, "Ctrl" is interpreted as "Command", so if you actually need "Ctrl", specify "MacCtrl".

So this ends up registering Cmd-C, which we obviously don't want.

If we remove the `mac` line, it's equivalent to Cmd-Shift-C, which, as for you, doesn't end up getting set by default for me. (That's the Inspector on macOS too, so if that's the reason, that would explain it.)

If we actually wanted to use Cmd-Ctrl-C, we'd need to use `Command+MacCtrl+C`. Another option for Mac users would be to just use `MacCtrl+C`. When I set up Word plugin shortcuts for a friend recently, I set it to Ctrl-C, and it's pretty darn convenient. (Word seems to bind Ctrl-C to copy by default for confused switchers, but you can override it.) But we probably don't want to steal Ctrl-C across all webpages.

On that note, in general I don't love the idea of reserving a shortcut key everywhere when we only use it on one domain. Most websites don't use shortcut keys, and the sites that do are unlikely to use complicated ones, but it's not impossible, and they could certainly use something like Ctrl-C. Would it be possible to pass through the shortcut key on other sites?

As for Windows/Linux, if we don't have reason to believe that it'll start using the default when it conflicts with a built-in shortcut (which [I think it should](https://forums.zotero.org/discussion/comment/326812/#Comment_326812)), we should use something different, since there's no reason to set a default that won't be used. But if we can't use Ctrl-Alt, I'm not sure what to use.

For all these reasons, it seems like the least-bad option may be to just configure this in the prefs where we already have a Google Docs setting.

PR #304, opened 2020-03-15
Use background page for all HTTP requests

dstillman:
Browsers have been progressively locking down HTTP requests from content
scripts, so we've moved requests to the background page when we know
they're cross-origin. Unfortunately, we can't detect ahead of time if a
request that looks like it will be same-origin redirects to another
domain, causing it to fail due to CORS/CORB [1]. While we could allow
translators to specify when a same-domain request might be redirected
across domains, it's easier to just do all HTTP requests via the
background page. I assume this is slightly less efficient, but it
probably doesn't matter in practice (and we exchange all sorts of
messages with the background page already).

(This addresses an issue reported on the dev list in a now-deleted
message about links such as http://theses.fr/2016ECDN0016/document that
can redirect to a different domain.)

@adomasven, let me know if you can think of any reason not to do this.

[1] https://www.chromium.org/Home/chromium-security/extension-content-script-fetches

### DIFF
--- a/src/common/http.js
+++ b/src/common/http.js
@@ -74,52 +74,35 @@ Zotero.HTTP = new function() {
 			successCodes: null
 		}, options);
 		
-		var useContentXHR = false;
-		
 		if (Zotero.isInject) {
-			// The privileged XHR that Firefox makes available to content scripts so that they
-			// can make cross-domain requests doesn't include the Referer header in requests [1],
-			// so sites that check for it don't work properly. As long as we're not making a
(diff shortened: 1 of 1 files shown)


adomasven:
I can imagine some of the requests translators perform might require Origin header of the translated site and there might be potential issues for Safari same-origin requests regarding cookies, since those are not automatically handled in the background page there and would need to be passed from the injected page.

adomasven:
Most requests are running via the background page, and also many changes to this code since for MV3, etc.

PR #306, opened 2020-03-23
Removed one comma and two comments from src\browserExt\manifest.json

pchemguy:
Build process uses jq processing, so comments should either be removed from the json input or filtered out. Additionally, comma on a line containing only a comment should be removed.


adomasven:
These are replaced during the build process with Gulp. See https://github.com/zotero/zotero-connectors/blob/master/gulpfile.js#L237-L243

PR #307, opened 2020-03-25
Fix CYGWIN/Gulp path issue in build.sh

pchemguy:
There is an apparent path related issue with Gulp when called via absolute path under CYGWIN. The issue does not occur if relative path is used.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -38,6 +38,10 @@ DONE
 }
 
 GULP=$CWD/node_modules/gulp/bin/gulp.js
+if [ "$IS_CYGWIN" == 1 ]; then
+	# Change Gulp path to relative for CYGWIN environment
+	GULP=./node_modules/gulp/bin/gulp.js
+fi
 
 BUILD_BROWSER_EXT=0
 BUILD_SAFARI=0
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -173,8 +173,11 @@ function replaceScriptsHTML(string, match, scripts) {
 function processFile() {
 	return through.obj(async function(file, enc, cb) {
 		console.log(file.path.slice(file.cwd.length));
-		var offset = file.cwd.split('/').length;
-		var parts = file.path.split('/');
+		// Under CYGWIN environment, the paths are converted to Windows style with "\".
+		// Additional ".replace(/\\/g, '/')" step leaves Linux-style paths as is,
+		// while "\" in Windows-style paths are replaced with "/" for further processing.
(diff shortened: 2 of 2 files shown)


dstillman:
Any patch needs to include a detailed justification — it's not enough to say "an apparent path related issue". Otherwise we're just accumulating cruft that could never be removed.

What's the actual error you're receiving? If the current directory has a space in it, the current code would fail, unrelated to the platform, because the path isn't quoted. Doing that would be a better fix.

pchemguy:
While in ./zotero-connectors, if I run
`/cygdrive/c/Users/pcuser/WebstormProjects/zotero/zotero-connectors/node_modules/gulp/bin/gulp.js -v`
it throws
> internal/modules/cjs/loader.js:979
>   throw err;
>   ^
> 
> Error: Cannot find module '**C:\cygdrive\c**\Users\pcuser\WebstormProjects\zotero\zotero-connectors\node_modules\gulp\bin\gulp.js'
>     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:976:15)
>     at Function.Module._load (internal/modules/cjs/loader.js:859:27)
>     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
>     at internal/main/run_main_module.js:17:47 {
>   code: 'MODULE_NOT_FOUND',
>   requireStack: []
> }

If I run
` ./node_modules/gulp/bin/gulp.js -v`
it produces the expected output.

watercrossing (commented):
Both commits are required to let build.sh succeed on cygwin on windows.

fbennett:
Hoping not to add to confusion ... two of us have worked through this over at Jurism. We wanted to have smooth connector builds running in Linux and in Windows, and here's what we've come up with. This differs a bit from this PR, but addresses the same issues. We're happy in our fork, but if it will help we can either put up a separate PR or leave it with the description below.

When building under Cygwin, the copy of node used for the build is installed under Windows (it isn't a Cygwin application). Given a relative path at the command line, it will extend its native Windows PWD, but if given an absolute path, it's passed through verbatim, and if that's a Unix path, things will break. To avoid relying on implicit magic inside the Windows node binary, the absolute Cygwin path can be specified with ``"$(pwd | xargs cygpath -w)``.

In the patch to ``gulpfile.js``, node's ``path`` module gives us delimiter arbitration for free, so with ``const pth = require('path');`` at the top of the file, the forward slashes at the top of ``processFile`` can just be replaced with ``pth.sep``, and the files will be processed okay.

PR #308, opened 2020-03-28
add a Filter to ItemSelector

fzyz999:
Hello, I found that it is useful if there are a filter in ItemSelector window. When you do literature research, you may want to download all papers related to some keywords from some conference or journal. For example, somebody may click a conference page(https://ieeexplore.ieee.org/xpl/conhome/8966450/proceeding), and download some paper whose title contains the keyword 'memory'. It is  inconvenience to do that manually. However, the ItemSelector doesn't provide any filter or similar features.

### DIFF
--- a/src/common/itemSelector/itemSelector.css
+++ b/src/common/itemSelector/itemSelector.css
@@ -16,7 +16,7 @@ body {
 	top: 20.5pt;
 	left: 2pt;
 	right: 2pt;
-	bottom: 32pt;
+	bottom: 64pt;
 	overflow: auto;
 	padding: 2pt 1.5pt;
 	border: 1pt solid black;
--- a/src/common/itemSelector/itemSelector.html
+++ b/src/common/itemSelector/itemSelector.html
@@ -26,9 +26,18 @@
 		<div id="label">Select which items you'd like to add to your library:</div>
 		<div id="item-selector"></div>
 		<div id="buttons">
+			<div id="filter">
+				<label for="reverseSelect" style="word-wrap:break-word">
+					<input type="checkbox" id="reverseSelect">Reverse Select
+				</label>
+				<label for="regExp" style="word-wrap:break-word">
+					<input type="checkbox" id="regExp">RegExp
+				</label>
+				<input type="text" id="pattern" placeholder="Pattern">
+			</div>
 			<div id="select-deselect">
-				<input type="submit" id="select" value="Select All"/>
-				<input type="submit" id="deselect" value="Deselect All"/>
+				<input type="submit" id="select" value="Select Items"/>
+				<input type="submit" id="deselect" value="Deselect Items"/>
 			</div>
 			<div id="cancel-ok">
 				<input type="submit" value="Cancel" id="cancel"/>
--- a/src/common/itemSelector/itemSelector.js
+++ b/src/common/itemSelector/itemSelector.js
@@ -44,6 +44,7 @@ function load() {
 	
 	// add checkboxes to selector
 	for(var i in items) {
+		console.log(i)
 		var title, checked = false;
 		if(items[i] && typeof(items[i]) == "object" && items[i].title !== undefined) {
 			title = items[i].title;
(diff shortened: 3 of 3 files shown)


dstillman:
Hi there — thanks for the PR.

I'm not sure the IEEE Xplore example is the most compelling use case, since you can filter the results further on the page itself, which would be even quicker because it fits more relevant results on the page at once. So this would really only be useful for saving multiple results from a search results page that for some reason can't do better filtering on its own, and I'm not sure how common that is. Even the more common use case of quickly finding something from the search results page that you want to save can be done already just by using the Cmd-F/Ctrl-F in the box.

If we did accept something like this, I think it would need to work differently, with just a simple text box that filtered the results as you typed, similar to the filter field in the tag selector in Zotero. Once filtered, Select All/Deselect All would apply to just the visible items, though hidden lines would keep their existing selection states. This would be pretty simple to implement — basically putting [this](https://github.com/zotero/zotero-connectors/blob/6eff330feea2037b91a225b3dd853d0dd3fec11d/src/common/itemSelector/itemSelector.js#L46-L76) in a function and calling it with the filtered list item on keydown. We wouldn't expose regexp or reverse select options.

We could test out something like that, but I'm just not sure it's necessary.

fzyz999:
Regexp and Reverse selection may be useless. I think you are right, these options shouldn't be exposed. However, I think a design like the filter field in the tag selector in Zotero is not enough. When I began to implement this filter, I have considered it. Because when we do literature research, we usually have multiple keywords. For example, someone may want to download all of the articles whose title contains 'cache' or 'memory'. If you just implement a filter like the filter filed in the tag selector, you can not select articles whose title contains both 'cache' and 'memory' at the same time(as you can only input one keyword). I'll adjust my PR as you suggested.

fzyz999:
Hi, I find another use case. Sometimes, you may need to deselect some useless items. For example, when you want to download all articles from https://ieeexplore.ieee.org/xpl/conhome/8401306/proceeding, you'll find some items like `ISCA 2018 Program Committee` or `ISCA 2018 Program Committee` which is meaningless. Another example is  https://dl.acm.org/toc/sigarch/2017/45/1, it contains several `Session details` items which someone may want to deselect.

fzyz999:
Hi, I have removed `Reverse Selection` and `Regexp` as you suggested.

dstillman:
> If you just implement a filter like the filter filed in the tag selector, you can not select articles whose title contains both 'cache' and 'memory' at the same time(as you can only input one keyword).

How is there a difference in filtering between what you implemented and what I'm suggesting? I'm just saying it should filter the list visually as you type, and Select All/Deselect All should apply to only visible items, as opposed to entering a pattern that then gets applied in a filtered way to the full visible list. They're both just matching against the strings.

The difference would just be treating the input as a phrase vs. splitting it up by whitespace and filtering for all entries that match all words. That could be done with either approach.

(Again, though, I would think you would normally just search for those keywords on the site itself before saving rather than doing complicated filtering in the Select Items box.)

> For example, when you want to download all articles from https://ieeexplore.ieee.org/xpl/conhome/8401306/proceeding, you'll find some items like ISCA 2018 Program Committee or ISCA 2018 Program Committee which is meaningless.

I don't see how using filtering there could possibly be faster than just unchecking a few boxes.

> Another example is https://dl.acm.org/toc/sigarch/2017/45/1, it contains several Session details items which someone may want to deselect.

I'm not getting the folder icon there (is that in a translator update that's not yet out?) but if that shows all the entries in all the sections it does seem useful for that one.

fzyz999:
Hi, thanks for your reply. I have reimplemented it as you suggested. 

> How is there a difference in filtering between what you implemented and what I'm suggesting? I'm just saying it should filter the list visually as you type, and Select All/Deselect All should apply to only visible items, as opposed to entering a pattern that then gets applied in a filtered way to the full visible list. They're both just matching against the strings. 

Sorry, I misunderstood your suggestion. You said "This would be pretty simple to implement — basically putting this in a function and calling it with the filtered list item on keydown." but I think the implementation you said is limited. I finally understand that you just saying it should filter the list visually as you type. That's a great suggestion. I agree with you. So I reimplemented it as you suggested.

> 
> (Again, though, I would think you would normally just search for those keywords on the site itself before saving rather than doing complicated filtering in the Select Items box.)
> 
You are right. In most cases, using the search function of the site itself is a more suitable > 
> I don't see how using filtering there could possibly be faster than just unchecking a few boxes.
> 
> > Another example is https://dl.acm.org/toc/sigarch/2017/45/1, it contains several Session details items which someone may want to deselect.
> 
> I'm not getting the folder icon there (is that in a translator update that's not yet out?) but if that shows all the entries in all the sections it does seem useful for that one.

PR #309, opened 2020-04-20
Fixes #289, PDF save failure in new tabs in Chrome

### LINKED ISSUE #289: PDF save failure in new tabs in Chrome
https://www.ajronline.org/doi/abs/10.2214/ajr.157.2.1853823

Click the PDF link in Chrome and try to save in the new tab.

```
messaging.js:121 Uncaught (in promise) TypeError: Cannot read property 'saveAsWebpage' of undefined
    at Object.saveAsWebpage (inject/inject.js:558)
    at messageListener (messaging_inject.js:119)
    at messaging_inject.js:129
```

Cmd-clicking the Bitcoin PDF from the Google search results and trying to save fails too.

It works after reloading the PDF, or when following a normal PDF link that doesn't open in a new tab.

dstillman:
I don't know why content scripts seem to be getting injected more than
once on PDFs loaded in new tabs in Chrome, but this seems to fix the
breakage that results.

### DIFF
--- a/src/common/inject/inject.jsx
+++ b/src/common/inject/inject.jsx
@@ -591,8 +591,8 @@ try {
 	isHiddenIFrame = !isTopWindow && window.frameElement && window.frameElement.style.display === "none";
 } catch(e) {}
 
-const isWeb = window.location.protocol === "http:" || window.location.protocol === "https:";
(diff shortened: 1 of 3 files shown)


dstillman:
It would be good to figure out why this is happening in Chrome, but this change obviously isn't going to break anything on its own, so I've merged it.

PR #319, opened 2020-07-17
Use SingleFile to create snapshots of webpages

fletcherhaz:


src/common/translate_item.js
```diff
@@ -93,7 +93,62 @@ Zotero.Translate.ItemSaver.prototype = {
...
+				if (items[x].attachments[y].mimeType !== 'text/html') {
+					continue;
+				}
+
+				// If we have not done so already, inject SingleFile and retrieve
+				// the page content
+				if (!pageContent) {
+
+					// Close the window so that we don't include it in SingleFile capture
+					Zotero.Messaging.sendMessage("progressWindow.close", null);
```
adomasven, line 115: I think this won't do. I was saving https://www.nytimes.com/2020/07/22/world/coronavirus-covid-19.html and upon clicking Save (with Embedded Metadata) the progress window appears very shortly, then disappears for about 10s and then reappears again, and this is on a fairly speedy CPU and with plenty of RAM to go around. Does Single-File not provide other mechanisms to hide certain elements by CSS selectors or similar?

dstillman:
After reviewing various options, the plan here is to try using a modified version of SingleFileZ to save a ZIP file — not its custom HTML+JS+ZIP format, which requires embedding JS processing code in every file and can't be viewed by default in non-Firefox browsers, but just the SingleFileZ folder structure in a regular ZIP file with a .zoterosnapshot extension that's displayed within Zotero but can always be unzipped if you want to access it externally.

dstillman:
Possible I didn't set up something correctly in the repo, but I couldn't actually get this to work. I checked out the other PR for the Zotero submodule and then tried to save the zotero.org homepage:

```
(3)(+0004176): saveAsWebpage message received in injected page https://www.zotero.org/
debug.js:171 (3)(+0000001): HTTP POST http://127.0.0.1:23119/connector/ping
debug.js:171 (3)(+0000005): Connector: Method ping succeeded
debug.js:171 (3)(+0000003): SingleFile: injecting SingleFile into page
debug.js:171 (3)(+0000001): SingleFile: injecting SingleFile into page
debug.js:171 (3)(+0000000): progressWindow.show message received in injected page https://www.zotero.org/
debug.js:171 (3)(+0000004): progressWindow.itemProgress message received in injected page https://www.zotero.org/
debug.js:171 (3)(+0000035): Inject: Injecting scripts into 0 - https://www.zotero.org/ : singlefile.js
debug.js:171 (3)(+0000065): Inject: Complete 0 - https://www.zotero.org/
debug.js:171 (3)(+0000001): progressWindowIframe.registered message received in injected page https://www.zotero.org/
debug.js:171 (3)(+0000001): HTTP POST http://127.0.0.1:23119/connector/getSelectedCollection
debug.js:171 (3)(+0000001): SingleFile: Retrieving page data
debug.js:171 (3)(+0000002): SingleFile: Error retrieving page data
debug.js:171 (3)(+0000001): {}
debug.js:171 (3)(+0000000): Fail retrieving page data
```

The `{}` is actually "TypeError: Cannot read property 'content' of undefined".

Also, in the page console, I get this:

```
Uncaught (in promise) TypeError: Cannot destructure property 'pageData' of '(intermediate value)' as it is undefined.
    at Object._saveAsWebpage (inject.js:496)
    at async Object.saveAsWebpage (inject.js:462)
```

This caused the whole save to hang, so we should make sure that doesn't happen even if there's a problem saving a snapshot. So 1) `retrievePageData()` should probably throw an error instead of swallowing it in the `catch`, so it's the real error instead of a destructuring-assignment error and 2) in `_saveAsWebpage()`, you can see an example of handling an unexpected error and sending a message to `progressWindow.done` in the `catch` further down.

src/common/translate_item.js
```diff
@@ -105,8 +105,52 @@ Zotero.Translate.ItemSaver.prototype = {
 			payload.cookie = document.cookie;
 		}
 		payload.proxy = this._proxy && this._proxy.toJSON();
+		
+		// Add page data for snapshot
+		// TODO: Check for skipSnapshot
+		let singleFile;
+		for (let i = 0; i < items.length; i++) {
+			if (items[i].itemType !== 'webpage') {
```
dstillman, line 113: It doesn't need to be a 'webpage' item for there to be a snapshot. Any item type can have one. See, e.g., ACS Publications — the tests show a 'bookSection' item with a snapshot.

src/common/translate_item.js
```diff
@@ -105,8 +105,52 @@ Zotero.Translate.ItemSaver.prototype = {
...
+				continue;
+			}
+
+			for (let j = 0; j < items[i].attachments.length; j++) {
+				if (items[i].attachments[j].mimeType !== 'text/html') {
```
dstillman, line 118: A snapshot is indicated by `mimeType: 'text/html'` and `snapshot !== false`. (`snapshot: false` is a linked-URL attachment. If `snapshot` is missing, that implies `true`.)

src/common/translate_item.js
```diff
@@ -105,8 +105,52 @@ Zotero.Translate.ItemSaver.prototype = {
...
+			for (let j = 0; j < items[i].attachments.length; j++) {
+				if (items[i].attachments[j].mimeType !== 'text/html'
+					|| items[i].attachments[j].snapshot === false) {
+					continue;
+				}
+
+				// If we have not done so already, inject SingleFile and retrieve
+				// the page content
+				if (!singleFile) {
+					singleFile = await Zotero.SingleFile.retrievePageData();
```
dstillman, line 122: This can't be the same logic as saveSnapshot, because in this case there's no guarantee that it's the current page. We only want to retrieve page data when the attachment URL matches the current URL.

Also, this can now only throw, not return undefined, so, as above, this should be in a try/catch and fall back to saving in the client if possible.

src/common/inject/inject.jsx
```diff
@@ -524,8 +517,42 @@ Zotero.Inject = new function() {
...
+			try {
+				let { pageData, form } = await Zotero.SingleFile.retrievePageData();
+				data["pageData"] = pageData;
+				form["payload"] = JSON.stringify(data);
+				data = form;
+			}
+			catch (e) {
+				Zotero.debug("Fail retrieving page data");
+				Zotero.Messaging.sendMessage("progressWindow.done", [false, 'unexpectedError']);
+				throw e;
```
dstillman, line 531: Actually, I guess I take back my previous comment — SingleFile failing shouldn't cause the whole save to fail. It'd be better to fall back to saving the attachment in the client, if we can do so easily.

src/common/translate_item.js
fletcherhaz: Ok. I think I had the URL comparison originally, but it was breaking on twitter because the query param was removed. Is that the only modification the attachment URL will have? I guess I could match on either the full or without the query params. Are those URLs coming from the translator?

src/common/inject/inject.jsx
fletcherhaz: Ok. That's still the the one piece I haven't upgrade to SingleFileZ. I was going to work on that today.

src/browserExt/singlefile.js
```diff
@@ -0,0 +1,147 @@
...
+
+			Zotero.debug("SingleFile: Done encoding page data");
+			return {
+				pageData,
+				form
+			};
+		} catch (e) {
+			// This attachment will fall back to SingleFile retrieval in the client/server
+			Zotero.debug("SingleFile: Error retrieving page data");
+			Zotero.debug(e);
```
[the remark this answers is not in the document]
fletcherhaz: I'm not sure why zotero.org is not working for you. It works just fine for me. Do any other webpages work for you?

src/common/translate_item.js
```diff
@@ -105,8 +105,52 @@ Zotero.Translate.ItemSaver.prototype = {
 			payload.cookie = document.cookie;
 		}
 		payload.proxy = this._proxy && this._proxy.toJSON();
+		
+		// Add page data for snapshot
+		// TODO: Check for skipSnapshot
+		let singleFile;
+		for (let i = 0; i < items.length; i++) {
+			for (let j = 0; j < items[i].attachments.length; j++) {
```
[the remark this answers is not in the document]
fletcherhaz: Ah ok. I didn't realize `for... of` is references for objects, but not for strings. Thanks!

src/browserExt/singlefile.js
```diff
@@ -138,10 +138,8 @@ Zotero.SingleFile = new function () {
 				form
 			};
 		} catch (e) {
-			// This attachment will fall back to SingleFile retrieval in the client/server
+			// This will fall back to SingleFile retrieval in the client
 			Zotero.debug("SingleFile: Error retrieving page data");
-			Zotero.debug(e);
-			throw e;
```
dstillman: Sorry, I meant only the debug line should be removed. It should still throw on error.

src/common/translate_item.js
dstillman: Hmm, this is tricky. OK, so to review the problem: Previously, the translator assigned a URL for each attachment, and the client would just retrieve that URL (with passed cookies) and save. We now want to save the current page in the connector if possible, but there's not really a great way to know if a current query string is meaningful and it should match the attachment URL.

We probably want to say that if 1) there's a single item being saved and 2) the full attachment URL matches some variation of the current URL, we assume the current page is the snapshot. Easy enough to match on full current URL or current URL with query string and fragment removed, but technically if the attachment URL has a query string we should match regardless of query parameter order, which would mean normalizing both the attachment URL and current URL to alphabetize any query parameters and remove any fragment.

What was the query param you were seeing on Twitter?

We should still fall back to the client for saving non-matching URLs (including for most multiples), so if our logic is imperfect here and we save the current page via the client with cookies, that's OK (though it won't work saving to server).

fletcherhaz: I think everything you said makes perfect sense! The query param I saw was `lang=en`, on twitter and probably found via Google. My biggest worry here is saving files when we are not supposed to.

I went back and double checked what was going on. I was saving a twitter page using Embedded Metadata and currently the ItemSaver comparing the `item.url` to `attachment.url`. I was incorrect though. `item.url` is coming across without the query parameter, but the `attachment.url` has the query param, at least in this example.

But it's not too complicated to build in the logic you suggested and I can't see a way that will go wrong so I'll go ahead and do that. And I'll compare against `location.href` not `item.url`.

src/common/translate_item.js
```diff
@@ -93,7 +93,62 @@ Zotero.Translate.ItemSaver.prototype = {
...
+			}
+
+			for (let y = 0; y < items[x].attachments.length; y++) {
+				if (items[x].attachments[y].mimeType !== 'text/html') {
+					continue;
+				}
+
+				// If we have not done so already, inject SingleFile and retrieve
+				// the page content
+				if (!pageContent) {
```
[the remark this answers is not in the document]
fletcherhaz: This was sort of old logic. I was going off of a multiple translator that saved a snapshot for every multiple item. This prevented us from redoing the SingleFile retrieval every time. However, that was bad logic on the part of the translator so it's no longer particularly relevant, but just in case, it speeds things up dramatically.

Also, the SingleFile logic has mostly been extracted out now.

gulpfile.js
```diff
@@ -269,6 +286,57 @@ function processFile() {
...
+				file.contents = Buffer.from(file.contents.toString()
+					.replace('const basePath = "../../../";', 'const basePath = "lib/SingleFile/";')
+					// .replace('const contentScriptFiles = [', 'const contentScriptFiles = [' + injects)
+				);
+
+				// Override the type so we include this file in firefox and chrome builds
+				type = 'browserExt';
+				// Switch from resource to lib sub-directory
+				parts[i+1] = 'lib';
+				break;
```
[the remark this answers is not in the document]
fletcherhaz: The reason I did this was that I was trying to be aware of what upgrading SingleFile would look like. By using their built in list of files to inject, I figured a new file would not slip through the cracks as easily. I also couldn't find a way to modify the files before injecting them so modifying the source on build seemed the right thing to do.

build.sh
```diff
@@ -284,6 +284,12 @@ function copyResources {
 	
 	# Remove .jsx files - we'll deal with those in gulp
 	find "$browser_builddir" -type f -name "*.jsx" -delete
+
+	# Copy SingleFile submodule code
+	mkdir -p "$browser_builddir/lib/SingleFile/extension/lib"
+	cp -r "$SRCDIR/zotero/resource/SingleFile/extension/lib/single-file" \
+		"$browser_builddir/lib/SingleFile/extension/lib/single-file"
+	cp -r "$SRCDIR/zotero/resource/SingleFile/lib" "$browser_builddir/lib/SingleFile/lib"
```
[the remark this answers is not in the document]
fletcherhaz: The multiple submodules was my mistake. It is only included once now. The reason for maintaining the directory structure was because I was using their injection script for the reasons described below. If we use our own list of files then yes we can absolutely change the name to something shorter.

dstillman:
If possible, we should do this in such a way that older clients will save snapshots normally without SingleFile — in other words, send the same info and just ignore it in the versions that support pageData.

Otherwise we'd need to check the client version and conditionally send pageData, but it'd be better to avoid that.

src/browserExt/singlefile.js
```diff
@@ -0,0 +1,139 @@
...
+		saveCreatedBookmarks: false,
+		ignoredBookmarkFolders: [],
+		replaceBookmarkURL: true,
+		saveFavicon: true,
+		includeBOM: false
+	};
+
+	this.retrievePageData = async function() {
+		// Call to background script to inject SingleFile
+		await Zotero.Connector_Browser.injectSingleFile();
```
dstillman, line 91: Let's put this within the try/catch, since calling code is swallowing any errors without logging.

PR #379, opened 2022-05-08
Fix `requestDocument` returning documents without locations, enable preprint support

AbeJellinek:
Type schema data needed to be updated so that `fieldIsValidForType` works with `preprint`, which enables `preprint` support in translators that use that method to test for it.

I'm not very familiar with connector dev/testing, but all tests that passed before (all but one, that is) still pass and translation works as expected.


dstillman:
Now available in the <a href="https://www.zotero.org/support/dev_builds#zotero_connector_beta">Zotero Connector beta</a>

PR #408, opened 2022-11-02
`Zotero.HTTP.request()`: Support 'json' responseType, return URL for `wrapDocument()`

AbeJellinek:
So apparently `request()` in the Connector didn't support returning JSON, which has been breaking our very few translators that call `requestJSON()`. Go figure! I probably should've caught that before merging in the async request functions. We also weren't returning `responseURL`, which `requestDocument()` uses to construct a wrapped Document.

(Some of these may be MV3 regressions, not sure, but I'm seeing them in the latest stable version on Firefox.)


dstillman:
Pushed to 5.0.102beta1 for Firefox

PR #417, opened 2022-12-29
Fix advanced-checkbox-enable-at-startup

yfdyh000:
Surprisingly, this option never takes effect.
I did not create a forum post to ask, and I was not sure if it was necessary. Hope this is not a mistake.

### DIFF
--- a/src/common/preferences/preferences.jsx
+++ b/src/common/preferences/preferences.jsx
@@ -274,6 +274,8 @@ Zotero_Preferences.Advanced = {
 			document.getElementById('advanced-checkbox-enable-logging').checked = !!status;
 		});
 		Zotero.Prefs.getAsync("debug.store").then(function(status) {
+			document.getElementById("advanced-checkbox-enable-logging").checked = !!status;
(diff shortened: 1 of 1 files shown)


adomasven:
This is intended to work the same way as the Zotero client's Restart with Debug Logging enabled option and shouldn't be reenabled on restart. But it indeed wasn't functioning. To be truthful I don't recall there being a bug that required a user to use this feature in the last 8 years, so that's probably how long this bug has been around for too, but I'm sure we will need it as soon as we decide to remove it since it's not being used.

yfdyh000:
I need this option, so I notice it. When developing a translator, I sometimes prefer frequently to restart the Connector extension to load the local updated translator instead of entering the options to update translators. Allow the logs in default may avoid missing anything or start it manually every time.

adomasven:
For development you can just open the background page inspector, where all debug lines are logged in the developer console. This logging flag is used only for storing the logs for submission to Zotero servers so that developers can troubleshoot bugs.

PR #460, opened 2024-02-19
Make the 'key' manifest prop env var configurable

adomasven:
See https://github.com/zotero/zotero-connectors/pull/456


gulpfile.js
```diff
@@ -295,6 +295,9 @@ function processFile() {
 							.replace("/*INJECT SCRIPTS*/",
 								injectScripts.map((s) => `"${s}"`).join(',\n\t\t\t'))
 							.replace(/"version": "[^"]*"/, '"version": "' + argv.connectorVersion + '"');
+						if (process.env.CHROME_EXTENSION_ID && ['chrome', 'manifestv3'].includes(browser)) {
```
dstillman, line 298: Did you mean `CHROME_EXTENSION_KEY` (which you used in the other repos)?

adomasven: Well noticed

PR #467, opened 2024-04-29
vpat 39: add aria labels to expand/collapse target button

abaevbog:
Add aria-expanded status and aria-label to the button that displays collection selector.

---

VPAT 39 description:

In the Zotero desktop app's browser addon, there is a button to expand or collapse the menu to provide additional information to the user. When using a screen reader, this button element does not have an accessible name. It also lacks a state to indicate whether the menu is expanded or collapsed. 

Impact Statement: When operable elements lack appropriate roles and states, users relying on assistive technologies (AT) may miss out on important functionality. If a collapsed menu is not announced as a collapsed menu, users may not realize it is operable and can be activated to show its expanded form. These issues are particularly impactful for users with visual or cognitive disabilities. 

Expected Result: Operable elements have appropriate names, roles, and states conveyed to users by assistive technologies. 

Actual Result: The expand/collapse menu button does not have a role or state conveyed to users by assistive technologies.


abaevbog:
Is better addressed in #473 among other things

Issue #469, opened 2024-04-29
vpat 38: citation successfully saved confirmation?


abaevbog:
VPAT 38:

Description: When using the browser extension to save citations, there is no alert or non-visual indication that the citation has been successfully saved. This lack of confirmation can lead users to unintentionally duplicate citations, as they are unaware that the first attempt to save was successful. 

Impact Statement: Users relying on assistive technologies may become confused when citations are duplicated due to the absence of a notification confirming that a citation was successfully saved. An alert that can be read as text and dismissed through user interaction would help alleviate this issue. 

Expected Result: An alert appears visually and phonetically, informing the user that a citation was saved to the library successfully. 

Actual Result: Without confirmation, the user will likely attempt to save the citation again, resulting in duplicate entries with each additional save attempt.

---

I'm not sure what we could do about this. When the connector appears, I believe the item is essentially already saved, so the "Saving to ..." part is in itself a confirmation. In addition, unless I reload the page, opening the connector multiple times (and even hitting "Done" button repeatedly) does not lead to duplicate entries. I am not sure I see how adding a success message would makes sense here. @adomasven, correct me if I'm wrong on this and there is an appropriate way to fit some kind of confirmation here.

dstillman:
Yes, this just seems like a misunderstanding. There's a success popup (which also allows changing the collection, but that's secondary), and it won't save again without a page reload.

The one thing I would want to check is whether we're assign a proper role (ARIA `alert` role?) to the save popup, so that it gets announced by screen readers.

PR #470, opened 2024-04-30
add dialog role to save item popup

abaevbog:
With the "Save to zotero" aria-label, screen readers should announce it when focus enters or leaves the dialog.

Addresses https://github.com/zotero/zotero-connectors/issues/469#issuecomment-2084448366

### DIFF
--- a/src/common/ui/ProgressWindow.jsx
+++ b/src/common/ui/ProgressWindow.jsx
@@ -62,7 +62,8 @@ Zotero.UI.ProgressWindow = class ProgressWindow extends React.PureComponent {
 			more: Zotero.getString('general_more'),
 			done: Zotero.getString('general_done'),
 			tagsPlaceholder: Zotero.getString('progressWindow_tagPlaceholder'),
-			filterPlaceholder: Zotero.getString('progressWindow_filterPlaceholder')
+			filterPlaceholder: Zotero.getString('progressWindow_filterPlaceholder'),
+			saveToZotero: Zotero.getString('progressWindow_saveToZotero')
 		};
 		
 		this.expandedRowsCache = {};
@@ -118,6 +119,8 @@ Zotero.UI.ProgressWindow = class ProgressWindow extends React.PureComponent {
 		(new Image()).src = 'disclosure-open.svg';
 		
 		this.sendMessage('registered');
+		
+		document.querySelector("#progress-window").setAttribute("aria-label", this.text.saveToZotero);
 	}
 	
 	
--- a/src/common/progressWindow/progressWindow.html
+++ b/src/common/progressWindow/progressWindow.html
@@ -30,6 +30,6 @@
 		<link rel="stylesheet" href="progressWindow.css" />
 	</head>
 	<body>
-		<div id="progress-window"></div>
+		<div id="progress-window" role="dialog"></div>
 	</body>
 </html>
--- a/src/messages.json
+++ b/src/messages.json
@@ -48,6 +48,9 @@
 	"progressWindow_savingTo": {
 		"message": "Saving to"
 	},
+	"progressWindow_saveToZotero": {
+		"message": "Save to Zotero"
+	},
 	"progressWindow_tagPlaceholder": {
 		"message": "Tags (separated by commas)"
 	},


dstillman:
The label on the popup is fine, but nothing should be on a timeout here.

We want to announce the "Saving to [target]" line from the dialog as it appears, so you know where it's being saved and can adjust it if necessary.

And then items — potentially multiple — appear in the popup as they're saved, so those are the lines that need to be announced. We could maybe add a prefix ("Saved" or something) for clarity. And we might need to debounce things so items appearing in rapid succession don't interrupt each other's announcements.

We would ideally also handle attachments somehow, which can take longer to turn from gray (saving) to black (saved) or can turn red (error). Not sure how best to do that. I think when item rows are added we know if there's an attachment, so we could announce the attachments in progress with the line, and then somehow handle their completions.

abaevbog:
Gotcha! I went for the timeout to be sure that the target is set but it's much better to do it in `changeHeadline`. 

And with attachments, there is `updateProgress` that processes messages of how the items are being added/downloaded, so I thought we could just hook alerts up to that? Now, it should announce "Save to ...", followed by "Item saved" for the parent item, followed by "Download started" and "Download finished" for the attachment. All page I've tested on have only one item and one attachment (not sure where there can be more) but I assume if there are, they'll just go through the same `updateProgress` workflow.

If there are two events for the same item coming in roughly at the same time, we'll only announce the later one (e.g. if download happens very fast, most likely we'll only announce "Download finished" part). 

Finally, NVDA and JAWS work as expected, but voiceover for me sometimes may prefer the focused dropdown over alerts (if the focus is there) or it might not announce anything if the focus does not go inside of the iframe (may happen if popup opened without mouse click). In any case, this seems like just another voiceover oddity.

Here I have a [demo with NVDA](https://www.dropbox.com/scl/fi/jsm0g4y9wqpm6f4ixtrzv/nvda_alerts_save_item_popup.mov?rlkey=bm597dw71dtlgk2svxuxroynj&st=xalg79hu&dl=0) going through the alerts

src/common/ui/ProgressWindow.jsx
```diff
@@ -261,6 +289,50 @@ Zotero.UI.ProgressWindow = class ProgressWindow extends React.PureComponent {
...
+		// Get the first message and announce if it is not the same a the last alert (sanity check)
+		let { message, _ } = this.alertQueue[0];
+		if (alertNode.textContent != message) {
+			alertNode.textContent = message;
+		}
+		// Delete the element from the queue
+		this.alertQueue.shift();
+		// Wait a bit and then try to announce the next message
+		setTimeout(this.processMessageQueue, internal);
+	}
```
adomasven, line 334: There are clean ways to sequentialize/chain calls without infinite waiting loops. E.g. see https://github.com/zotero/zotero-connectors/blob/239745e644ae774192f8b390adbaaab30103a03c/src/common/inject/inject.jsx#L292-L315

adomasven:
Try https://arxiv.org/abs/2405.00757 for something with multiple items. It's good in terms of announcing what's happening, but maybe we need configurable delay time based on the length of the text? Also saving sometimes does fail, so we should test that and do something reasonable.

abaevbog:
> There are clean ways to sequentialize/chain calls without infinite waiting loops. E.g. see

Yeah, having looked at it again, I agree that repeated calls through setTimeout was unnecessarily convoluted. I kept the part with the original `alertQueue` because it allows us to remove alerts from the queue more easily but otherwise, I think it's nicer to just process them in a `while` loop with an `await` of a delay for the announcement inside of it. What do you think?

> It's good in terms of announcing what's happening, but maybe we need configurable delay time based on the length of the text?

I tried https://arxiv.org/abs/2405.00757 and everything gets announced to me as expected, so that's nice! With the variable delay ... this is very tricky. The actual speed of screen readers varies greatly so trying to estimate how long a sentence would take to announce, in my opinion, is somewhat impossible... No hurt to try to estimate it though based on the average speed of ~200 words per minute which gives us about 3.33 words per second but that actually leads to my question below.

The popup may close before all announcements are read out. In some cases, it may stop the announcements as they are half-way through. If it doesn't stop, it's weird if alerts are still going for a while after the popup closed (I've encountered both behaviors on the same setup with NVDA). So next my thought was to just make sure that each alert is short enough which increases our chances of not having this problem. At first I tried to have announcements like "Item X out of Y downloaded" - but that doesn't quite work because the download of, say, 3d item can get announced before the 2nd item, which is just confusing. So I tried to truncate the announcements to only have 5 words, which to me seems like the nicest solution. I think it's enough to identify the item while being brief. Does that seem reasonable? I also am a bit concerned about alerts going after the popup closing at all - that might be a bit disruptive? So maybe we should clear our all alerts once the popup closes...

For reference, this is the last [recording](https://www.dropbox.com/scl/fi/a56xloktu03w0fgcj3if3/nvda_multiple_items_popup.mov?rlkey=tysj7gbjf3bm2qagtou820zbz&st=lkoct7zn&dl=0) from https://arxiv.org/abs/2405.00757.

abaevbog:
> Also saving sometimes does fail, so we should test that and do something reasonable.

This should be taken care of by the status from `updateProgress`, and I also added an `alert` to errors container which announces the rendered errors at the bottom. There may be some variation with events being skipped or not depending on the sequence but [this](https://www.dropbox.com/scl/fi/kx1vxxb032z5r70l3yfty/errors_popup.mov?rlkey=630agm9ow1s1xnnk286dqwem3&st=3f78x5ub&dl=0) is what it comes out as for me if there is no internet

adomasven:
> The popup may close before all announcements are read out. In some cases, it may stop the announcements as they are half-way through. If it doesn't stop, it's weird if alerts are still going for a while after the popup closed (I've encountered both behaviors on the same setup with NVDA). So next my thought was to just make sure that each alert is short enough which increases our chances of not having this problem. At first I tried to have announcements like "Item X out of Y downloaded" - but that doesn't quite work because the download of, say, 3d item can get announced before the 2nd item, which is just confusing. So I tried to truncate the announcements to only have 5 words, which to me seems like the nicest solution. I think it's enough to identify the item while being brief. Does that seem reasonable?

That seems good. I only tested with VoiceOver, but it was cutting off alerts without fully reading them out anyway, so it was already doing pretty much the same thing.

A couple of thoughts.

1. The child item names depend on translators and some reasonable translation framework defaults when translators do not provide names. Still, sometimes it may not be clear what child item-types are being saved from those titles, so perhaps that should be indicated?
2. When saving multiple items (like from the arxiv feed page), it is not clear which are parent items and which are child items.

Given all that, maybe something along the lines of "Item X out of Y downloaded" is appropriate. It could be something like 

>Saving X items. Item 1 [Title{5}]. Item 2 [Title{5}] [...] Saved a comment for Item 1: [Comment{5}]. Saved a page snapshot for Item 1. Failed to save PDF for Item 1.

> I also am a bit concerned about alerts going after the popup closing at all - that might be a bit disruptive?

I don't know. For someone visually impaired does it even matter that the popup had closed? It does sound like there needs to be a way to stop the alerts from keep on being read if you had just saved 10 items and each is saving 3 child items. Is there some default gesture/command that screenreaders usually use for that which we could support?

abaevbog:
> Saving X items. Item 1 [Title{5}]. Item 2 [Title{5}] [...] Saved a comment for Item 1: [Comment{5}]. Saved a page snapshot for Item 1. Failed to save PDF for Item 1.

That is a good setup! I did some refactoring to try to keep track of that, and [this](https://www.dropbox.com/scl/fi/mocbc7ga2v5i2cgiwftx5/refactored_alerts.mov?rlkey=6nlkq0dijkinx4ghb8m27bb3t&st=x9oim42l&dl=0) is roughly what I am going after now?

> Is there some default gesture/command that screenreaders usually use for that which we could support?

It seems like a simple Ctrl keypress is a generally accepted "stop talking" button for all screen readers. From my testing, it does pretty much just that. So that concern of mine doesn't seem like a really big issue.

One unfortunate observation I just made is that constantly changing `alert` causes [issues](https://www.dropbox.com/scl/fi/c9d4jqpkld6xrhtghvck2/Jaws_alert.mov?rlkey=hlqi978xffcikt32o8025574r&st=5sbukpni&dl=0) with JAWS - unlike NVDA or voiceover, it keeps interrupting the previous statement adding "alert" before each notification. It leads to multiple "alert... alert .... alert ...", which is not good. So I'm working around that next - the notification area has to become `polite` to avoid it (role=`log` seems appropriate too), but ... this had voiceover duplicating announcements with this setup. So there are some things to reconcile!

adomasven:
> That is a good setup! I did some refactoring to try to keep track of that, and [this](https://www.dropbox.com/scl/fi/mocbc7ga2v5i2cgiwftx5/refactored_alerts.mov?rlkey=6nlkq0dijkinx4ghb8m27bb3t&st=x9oim42l&dl=0) is roughly what I am going after now?

1. Seems like the initial announcements are a bit too verbose? Not sure if terseness is considered a goal for screen readers? I think when saving 1 item it should just say "Saving 1 Item [Title{5}]. [Attachments...]". For multiple items it should be "Saving 2 Items. Item 1 [Title{5}]. Item 2 [Title{5}]. [Attachments...]"
2. For attachments it should say "Saved" instead of "Downloaded". "Downloaded" may be more appropriate for PDFs, but "saved" for everything is just cleaner.
3. We should not be announcing pdf and snapshot labels. Like in my example above, it should just say "Saved a page snapshot for Item 1".

> It seems like a simple Ctrl keypress is a generally accepted "stop talking" button for all screen readers. From my testing, it does pretty much just that. So that concern of mine doesn't seem like a really big issue.

Does it actually work if we keep on adding new stuff to the alert element? Do new additions not trigger re-announcement?

abaevbog:
> Seems like the initial announcements are a bit too verbose?

That's fair - in general, I think it's best to try to keep things short unless there is a reason not to. Here are the refactored [notifications](https://www.dropbox.com/scl/fi/0xs0ny6wxigolc2zb6dgc/refactored_notifications.mov?rlkey=9iqc3wqhqn91o1sh7fbwwl8ht&st=nu3pq4qs&dl=0) without child items' labels and condensed initial announcement. So now it basically goes
```
Saving X items. Item name 
Saved note for item 1
Saved PDF attachment for item 1
Saved Snapshot for item 1
```

For the same reason of reducing verbosity, I also removed the "Saving attachment" message for now. Having gone through the workflow of saving items many times, my experience was that it mainly fires right before the "Saved attachment" message. It leads to almost the same notification (except for the "ing" part) being announced twice. It gets especially annoying if there are many items, so instead of 5 messages you may get 10.

Another small refactoring here is that I made the role of the alerts container a [log](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/log_role). It's more appropriate semantically and [alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role#description) should be used sparingly (mainly for things that require immediate attention like confirmation dialogs and such). So now, instead of setting a message in the container, we're appending messages into it, and screen readers decide when to announce them based on their settings/other alerts. I did keep the first "Saving to ..." `alert` - this one seems important enough for that. 

> Does it actually work if we keep on adding new stuff to the alert element? Do new additions not trigger re-announcement?

It did work consistently for me but with different browsers and screen readers, it is a fair concern. I was trying it out after the popup was hidden (when the JS queue is more likely to have been emptied out), so that may be the reason why. Thinking about this was what made me drop our explicit JS queue and send all messages into a `role="log"` container. That way, once everything is loaded, all content is already appended into the log, so there is no reason for announcements to get re-triggered if they are stopped. It certainly works better as the items are being loaded compared to the previous commit. Before, if you want to stop announcements and press Ctrl, it would really do nothing because the current message will be interrupted but the next one would start right away. Now, the log container is updated less frequently, so if you press Ctrl as the items are being loaded, it will actually be quite for some time (until new announcements are generated after debounce).

One question I had was with the itemTypes and localization. I added a function `determineAttachmentType` to decide if it's a snapshot/pdf/etc. Is there a better way to do this? Are there more contentTypes I should include?

Current "progressWindow_savingItems" in messages.json is "Saving $1 items. $2", which is plural. Is there some kind of special localization process as with zotero.ftl to follow to account for cases when only one item is saved? Or do I just add another message "Saving $1 item" and decide which one to use in `handleAlerts`?

adomasven:
> One question I had was with the itemTypes and localization. I added a function determineAttachmentType to decide if it's a snapshot/pdf/etc. Is there a better way to do this? Are there more contentTypes I should include?

I don't think there is at the moment, but we will want to localize those strings too.

> "progressWindow_savingItems" in messages.json was "Saving $1 items. $2", which is plural. I just added another message "Saving $1 item" and decide which one to use in handleAlerts. Is there some kind of special localization process to follow instead of that?

I don't think there is support for that in the web extension framework (shame!).

abaevbog:
> I don't think there is, but we will want to localize those strings too.

Oh, right. I moved them all into `messages.json`.

> I don't think there is support for that in the web extension framework (shame!).

I know! The documentation is not very clear on it - pluralization is just nowhere mentioned. I did try but it seems like placeholder substitution is all we can do ... So yeah, we'll just have to have two separate entries for when there is one top level item vs multiple and pick the right one in `handleAlerts`.

adomasven:
Merged via e08f8ac39a66cb76c6c675b2d0a3e4b701850d97

Issue #472, opened 2024-05-10
Progress window accessibility localization tweaks


dstillman:
See comments on https://github.com/zotero/zotero-connectors/commit/e08f8ac39a66cb76c6c675b2d0a3e4b701850d97

PR #473, opened 2024-05-10
accessibility localization tweaks

abaevbog:
- Removed "Item 1" prefix if only one item is being saved. The new announcements will be just "Saving item {title}. Saved snapshot for this item"
- "Web link" instead of "Linked attachment" as the item type label
- Added epub content type to be announced as an attachment
- Use `general_saveTo` string for the label of the dialog
- Added aria-labels to the expand/collapse twisty button (vpat 39). I started with "View details"/"Hide details". Correct me if there's a more descriptive way than "details" to refer to the collections table and the tags area together.
- Added aria-label to the collection selector table so that it indicates what that element if for when a row is focused. Again, let me know if there is a better way to name this than "Target collection selector".
- Fixed a race condition bug to not trigger handle alerts until all top level items are 100 percent loaded.

Fixes: #472

### DIFF
--- a/src/common/inject/inject.jsx
+++ b/src/common/inject/inject.jsx
@@ -264,6 +264,7 @@ Zotero.Inject = new function() {
 		if (attachment.linkMode === "linked_url") return Zotero.getString("itemType_link");
 		var contentType = attachment.contentType || attachment.mimeType;
 		if (contentType == "application/pdf") return Zotero.getString("itemType_pdf");
+		if (contentType == "application/epub+zip") return Zotero.getString("itemType_epub");
 		if (contentType == "text/html") return Zotero.getString("itemType_snapshot");
 		return Zotero.getString("itemType_attachment");
 	}
--- a/src/common/inject/progressWindow_inject.js
+++ b/src/common/inject/progressWindow_inject.js
@@ -230,6 +230,7 @@ if (isTopWindow) {
 		var iframe = document.createElement('iframe');
 		iframe.id = frameID;
 		iframe.src = frameSrc;
+		iframe.title = Zotero.getString('general_saveTo', 'Zotero');
 		iframe.setAttribute('data-single-file-hidden-frame', '');
 		var style = {
 			position: 'fixed',
@@ -477,6 +478,7 @@ if (isTopWindow) {
 		}
 		else if (returnValue[0]) {
 			startCloseTimer(3000);
+			addEvent("willHide");
 		}
 		else {
 			if (returnValue.length < 2) {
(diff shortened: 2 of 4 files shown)


dstillman:
I haven't tested without this PR, so probably not from this specifically, but can we do anything about the backwards order here (collection before "Saving to")? Maybe it's correct, in the weird logic of screen readers, because it's telling you the current value before giving you the context?

Screenshot 2024-05-16 at 7 18 13 AM

Can we do anything to announce the dialog closing? I'm finding it keeping the VO selection in place even after the dialog closes, with no announcement that that happened.

abaevbog:
> I haven't tested without this PR, so probably not from this specifically, but can we do anything about the backwards order here (collection before "Saving to")? Maybe it's correct, in the weird logic of screen readers, because it's telling you the current value before giving you the context?

The exact way of how the elements' announcements are put together based on the DOM and aria properties is something that's specific for each screen reader. Voiceover is particularly odd about it - it always says value first and label after. On the contrary, JAWS and NVDA both say "Saving to combobox '{Selected collection}'" for that dropdown. So I don't think this is something we can or should try to change, since this way of announcing things seems specific to voiceover and users are likely accustomed to and expect it...

Somewhat related to this is that voiceover often decides to immediately announce the focused menupopup and skip the explicit "Saving to {collection}" [alert](https://github.com/zotero/zotero-connectors/blob/aa198b4a29d2335cebc9de59f95fd6e2a9189513/src/common/ui/ProgressWindow.jsx#L202) that fires right as soon as the popup opens. This would have helped to clarify things a bit. In general, I've noticed that voiceover may unfortunately skip some `aria-live` announcements depending on what else is going on.


> Can we do anything to announce the dialog closing? I'm finding it keeping the VO selection in place even after the dialog closes, with no announcement that that happened.

That is a great point! I added a new `progressWindowIframe.willHide` message that fires when all saving is done, so it will add an additional message before the dialog closes. Let me know if the phrasing should be tweaked. I went with "Saving is complete. The dialog will close now." as a start.

Also, I added a special alert that the items have already been saved on re-opening of the dialog. That way, it doesn't say "Saving to ..." because then it does make it seem like it's saving duplicate items, which is not true.

[New announcements video](https://www.dropbox.com/scl/fi/h3nvmum0cnq3kz0bwbu79/new_announcements.mov?rlkey=8x9ttoszlk8sg42ppd3rj9721&st=13e9tm3t&dl=0)

abaevbog:
A somewhat bigger issue I noticed with voiceover is that it does not announce anything at all if the focus/cursor is not in the actual web page. That means that if you tab into the "Save to zotero" extension button in browser's toolbar and press "Enter", nothing gets announced because (I think) the focus is on the toolbar. Since NVDA and JAWS do not have this issue, it seems like just another oddity of voiceover that I'll try to workaround somehow

adomasven:
Merging as this is an improvement over the current state and we want to release it and we can refine further if need be.

Issue #301, opened 2020-02-29
Cancel button in expanded save popup


dstillman:
Similar to Cancel/Remove Bookmark in Firefox if you click the star by mistake

twio142:
I'd also like to see this.

dstillman:
https://twitter.com/WaterPotential/status/1420755867113181187

XuShenLZ:
Hope this issue can be addressed soon. 

I feel like there should at least be an extra "confirm / save" button after we make sure we don't click on this extension by mistake, and we have specified the right collection to save the current ref.  Currently it automatically save to the default (last saved) collection and I have to manually move it if that's not the place I want.

dstillman:
> I have to manually move it if that's not the place I want

To be clear, that's the point of the Recents drop-down and collections list in the save popup. There's no expectation that the correct collection will be selected initially.

XuShenLZ:
OK got it. I think there were time that when I change the collection in the popup menu, it ends up creating 2 entries in both locations. But it didn't happen when I tried again just now.

Looking forward to the cancel button!

dstillman:
https://forums.zotero.org/discussion/117338/how-to-cancel-adding-item-from-the-web-browser-extension

PR #502, opened 2024-08-09
Update icons to Zotero 7 style

adomasven:
- Chrome, Firefox and Edge gets hi-DPI toolbar images
- Stop generating greyscale icons for Safari, since Safari now supports colored toolbar button images
- Generate pngs from svg for toolbar using rsvg-convert since toolbars do not support svgs
- Update Zotero submodule (for new icons)
- Update utilities submodule (for zoteroTypeSchemaData.json)

Closes #501

PR #512, opened 2024-09-18
add cancel button to undo unwanted save session

abaevbog:
- Added cancel button to undo saving of items in case it was started by mistake
- If none of `/connector/save*` endpoints were called yet, the button will just exit at saveItems.
- If the connector server has been engaged, hit `/connector/cancel` to tell the server to undo this session
- After cancel button is clicked, the connector window is hidden
- Minor style tweaks to fit the cancel buttons on the same row as tags input and done button

Depends on: https://github.com/zotero/zotero/pull/4694 that adds the `/connector/cancel` endpoint

Addresses: https://github.com/zotero/zotero-connectors/issues/301

A rough demo with the current state of `/connector/cancel` endpoint.

https://github.com/user-attachments/assets/0076bfdb-ae45-4e48-b3a1-943e83fb969f

### DIFF
--- a/src/common/progressWindow/progressWindow.css
+++ b/src/common/progressWindow/progressWindow.css
@@ -46,7 +46,7 @@ body {
 	background-color: var(--color-sidepane);
 	box-shadow: 0px 0px 3px 0px #0000008C, 0px 6px 10px 0px #00000040;
 	margin-left: 4px;
-	padding: 6px 10px 6px;
+	padding: 12px;
 	min-height: 40px;
 	overflow: hidden;
 	
@@ -74,7 +74,6 @@ body {
 	font-family: system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
 	font-size: 13px;
 	font-weight: bold;
-	margin-top: 2px;
 	margin-bottom: 8px;
 	white-space: nowrap;
 	text-overflow: ellipsis;
@@ -84,12 +83,12 @@ body {
 }
 
 .ProgressWindow-headlineSelect {
+	width: 240px;
 	margin-left: 5px;
 	height: 22px;
 	font-size: 13px;
 	overflow: hidden;
 	text-overflow: ellipsis;
-	background-color: white;
 	border-radius: 5px;
 	border: 1px solid var(--fill-quarternary);
 	background-color: var(--color-background);
@@ -263,7 +262,7 @@ body {
 
 .ProgressWindow-button.disclosure {
 	min-width: 22px;
-	margin-inline-start: auto;
+	margin-inline-start: 8px;
 	fill: currentColor;
 	color: var(--fill-primary);
 	background-color: var(--color-button);
@@ -347,7 +346,7 @@ input::placeholder, textarea::placeholder {
 
 .ProgressWindow-progressBox-separator {
 	border-top: 1px solid var(--fill-quarternary);
-	margin-inline: -10px;
+	margin-inline: -15px;
 	margin-bottom: 12px;
 	margin-top: 8px;
 }


abaevbog:
@adomasven I wanted to have a look at one of the connector issues while a number of accessibility changes is being reviewed and this seemed like a good one. Let me know what you think! This ended up being spread across two PRs - I marked both of them as drafts to begin with to make sure it's a reasonably direction to move in.

adomasven:
This is great!

I'm not a fan of the Done and Cancel button placement though, but that's something to address more generally perhaps, since even now Done is located somewhat randomly in the middle of the dialog. We cannot put buttons at the bottom as the dialog expands as items are saved. @yexingsha do you have any ideas?

yexingsha:
> We cannot put buttons at the bottom as the dialog expands as items are saved.

I'm not sure I understand the rationale behind this. A cancel button that doesn't appear without expanding the dialog first seems pretty counterintuitive. Is there a reason we can't put the buttons at the bottom of both expanded and unexpanded versions of this dialog?

abaevbog:
We can! I feel like "Done" and "Cancel" buttons should be together, otherwise the "Cancel" button would just be surrounded by empty space. So we could place both of them onto the next line and let tags input fully expand. Something like either of these options?

This still looks a bit strange to me because of the empty space after/before the buttons but the second option seems somewhat more conventional?

Screenshot 2024-09-18 at 10 19 27 AM

Screenshot 2024-09-18 at 10 21 38 AM

abaevbog:
> A cancel button that doesn't appear without expanding the dialog first seems pretty counterintuitive. Is there a reason we can't put the buttons at the bottom of both expanded and unexpanded versions of this dialog?

So I just realized that I did not fully understand the comment and just jumped to making suggestions with buttons still not visible unless the collection selector is expanded... Sorry about that! So my last comment is somewhat irrelevant those the screenshots are still possibilities to consider.

As far as I am aware, technically, there is nothing stopping us from showing the buttons at the very bottom of the dialog always. We could pull them out of the target selector component. Does it feels a bit strange because they end up being "pushed down" as the items are being added though?

https://github.com/user-attachments/assets/064232c9-6572-4089-8e2f-bd641d4922a2

And if we were to go this way, we'd want to do something with the "Done" button because it just hides the popup without stopping the save process, as opposed to confirming that "everything is finished". So in the recording above, it feels a bit misleading to me. Maybe we could disable it until we're really done or something?

yexingsha:
I think the cancel/done buttons being pushed down is fine, since it's just like e.g. macOS "Save As" dialog, where expanding the dialog pushes the buttons out and down.

As to the "Done" button being misleading, I think when people click this button, they probably just want to dismiss the dialog. The button being disabled for a few seconds will feel annoying and laggy, and anything longer than a few seconds will make people think something has gone wrong. Borrowing from the "Save As" dialog again, maybe we can change the wording to "Save" so that it's more like acknowledging an action rather than confirming the action has finished?

adomasven:
1. To give some context, initially the progress window did not have a collection picker and clicking anywhere on it dismissed it. This is still the case with the unexpanded dialog. I think it's generally fine and is better than adding a "Done" button, since people might start assuming they have to click that otherwise saving won't go through.

2. We added a the "Done" button to the expanded dialog since you have a collection selector and tag entry box there, and we felt we needed users to give something to click so that they can feel certain their changes have been saved, but technically the button just closes the progress window.

To maintain 1, it would probably be best to only add a cancel (X/trash/undo icon?) button somewhere on the unexpanded progress window. In neither case the buttons should be placed at the bottom of the dialog, because if you're saving multiple items, that dialog might keep expanding and also eventually gain a scrollbar (hiding the buttons below the fold). 

Also this may be a personal opinion, but it's true, that software which moves UI elements under user's direct cursor are committing a cardinal sin of UX and if some macOS dialog does that, then shame on them!

Also we should probably make the buttons blue and red.

yexingsha:
I think not adding a "Done" button to the unexpanded dialog is fine. And we should make the "Done" button in the expanded dialog blue, but making the "Cancel" button red is probably unnecessary, because it's not really a destructive action that warrants a warning.

As to the cancel button moving around, I think it's okay when it happens as the dialog gets expanded, because that's expected from a user action, and the user's cursor is on the expand/collapse button. The case where multiple items are being saved can indeed cause the cancel button to move away from directly under the user's cursor. However, I'm still inclined to keep the buttons at the bottom of the dialog because that's the standard pattern, and people will expect them to be there. We can do two things to alleviate the issues:

1. Put the button in a separate div so that when the content overflows, scrolling won't hide the buttons.

2. When there are multiple items being saved, set the height of the initial dialog to maximum to avoid the buttons moving around.

multiple-bottom-expanded
multiple-bottom-collapsed
single-bottom-collapsed

If we have to put the buttons somewhere else, the least weird place I can think of is at the top. That does avoid all of the issues, but still I don't really recommend it:

multiple-top-collapsed
multiple-top-expanded

abaevbog:
@adomasven thank you for the context! It actually clarified a few things for me. I've wondered for a while why there is a "done" button that essentially just closes the popup.

@yexingsha from the suggested ideas, the one where we place the buttons at the bottom and the progress items containers becomes scrollable if we used up all available height is the most appealing to me personally. While the buttons being pushed down does feel weird, I suppose anything else would be even weirder, right? Like, buttons at the top is very not-standard, and if we place them below the collection selector but above the progress items like [here](https://github.com/zotero/zotero-connectors/pull/512#issuecomment-2359032284), it'll also look strange because the "done" and "cancel" button would be in the middle of the popup. And having the popup immediately expand and leave empty space between buttons and the last progress item I think would look as if something broke and is missing there.

So I began looking at putting option 1 together. It does require a bit of stylesheet refactoring that I'm working through but this is the rough demo of this with a large number of items being added. Besides the tiny icons (that I'm looking into), I think it feels quite alright?


https://github.com/user-attachments/assets/ebbc76eb-933b-4af9-aca2-f3a3af5d983e

adomasven:
It's weird that the scrollbar doesn't appear for sections that are scrollable. I guess this is a macOS only thing, but can we force it to appear there? Also, if you accidentally saved something, I think you're going to notice it and try to click the Cancel button early, which means that the button moving away IS going to be annoying, and furthermore you will click on the dialog and dismiss it.

One thing we could do is make the dialog stop expanding if the user's cursor is hovering over it and then allow it to expand again if it leaves or expands the target selector. It might feel a bit strange, but we're going to have to make some sort of compromise here.

Another idea is to make the buttons floating or put them on the side like some weird appendix or something. Or make the dialog wider and put the cancel button on the first line and make it an X next to the down-chevron that expands into a Cancel button on hover.

abaevbog:
> It's weird that the scrollbar doesn't appear for sections that are scrollable. I guess this is a macOS only thing, but can we force it to appear there?

Hm, that comes from macOS system preferences > Appearance > Show scrollbars, and I'm not exactly sure if we can force it to show the scrollbars if the system pref is set otherwise... It's worth noting that selecting to always display a scrollbar looks like this. The extra outermost scrollbar is not meant to be there so it needs to be somehow removed.

Screenshot 2024-09-20 at 12 05 39 AM

>  Also, if you accidentally saved something, I think you're going to notice it and try to click the Cancel button early, which means that the button moving away IS going to be annoying, and furthermore you will click on the dialog and dismiss it.

That is a great point. You would be "chasing" the cancel button and that is not good ...

> One thing we could do is make the dialog stop expanding if the user's cursor is hovering over it and then allow it to expand again if it leaves or expands the target selector. It might feel a bit strange, but we're going to have to make some sort of compromise here.

I'm afraid the risk here would be that if you open a popup, hover over it immediately and do nothing, it might look like something broke because nothing would be happening ... We could add one or two items and then just scroll the container down as more items are added but it goes back to pushing the button initially and seems to complicated. 

> Another idea is to make the buttons floating or put them on the side like some weird appendix or something.

Alternative positioning for buttons is definitely something to consider. One thing with floating buttons to keep in mind is, per the vpat review issue, one needs to be able to zoom in very closely (400%) into the popup. So if the buttons are floating or sticky, we need to make sure the rest of the content can still be visible and scrollable. 

> Or make the dialog wider and put the cancel button on the first line and make it an X next to the down-chevron that expands into a Cancel button on hover.

If the dialog becomes winder, maybe we just put "Cancel" button (as an X with a tooltip)  and "Done button" together onto the very first line between the collection dropdown and the expand button? It could be somewhat close to what dropbox displays when you upload items:

Screenshot 2024-09-20 at 12 43 56 AM

Last idea I had is to put buttons below the target selector (as I tried first) but make a more distinct separation between the entire popup above and the progress items below. Add more padding, some border, etc. That way, there would be an interactive popup section that ends with "cancel" and "done" buttons and the progress items would be something like a log below. This idea came from looking at a github actions log, so it may or may not be appropriate for most people.

Screenshot 2024-09-20 at 12 24 24 AM

adomasven:
> Last idea I had is to put buttons below the target selector (as I tried first) but make a more distinct separation between the entire popup above and the progress items below. Add more padding, some border, etc. That way, there would be an interactive popup section that ends with "cancel" and "done" buttons and the progress items would be something like a log below. This idea came from looking at a github actions log, so it may or may not be appropriate for most people.

Yeah, and this doesn't work with the collapsed dialog.

abaevbog:
I was thinking something like below. It actually looks better to me when the dialog is collapsed than when expanded:


https://github.com/user-attachments/assets/37bd1b13-0f22-4026-9be1-12a610c89a07

And if there are too many items, that section would be scrollable

adomasven:
This could work with more event vertical padding and left-alignment although I still think it's worth exploring an X that expands into the word Cancel on Hover/Focus.

abaevbog:
> This could work with more event vertical padding and left-alignment

Ok, cool! I am really just brainstorming ideas trying to not get too stuck on specifics so all of the suggestions above would very much need to be polished.

> I still think it's worth exploring an X that expands into the word Cancel on Hover/Focus.

It is definitely a possibility to place the X into the header. Trying it out, I think we do run into a few small issues. 1. If we place the cancel button to the top, I feel like the "done" button needs to go there too? I think it would look very strange if two buttons that generally go side-by-side are in different parts of the popup. In that case, different widths of done, cancel and expand icons look make it look a bit strange. 2. If we also want the cancel button to expand from X to "cancel", we either have to reserve space for the expanded form, which means "X" will be surrounded by empty space, or it will push other buttons on hover, which is also strange. That is, unless the cancel button is the left-most button but I think the X button always goes to the very right or the very left.

This is a brief recording to demonstrate these points:

https://github.com/user-attachments/assets/212dc42e-821d-4457-8892-1b6625afdc96


As a potential remedy, what if we do not expand the "cancel" button and keep it as an "X" but add a `title="Cancel"` so that if someone is not sure what it does, they still get a hint when they hover over? Then, we'd have cancel and expand/collapse button as just icons, so maybe we make the "done" button a simple icon as well? Then all 3 icons are somewhat unified and it will look almost similar to typical window control buttons. This is a sketch of what it might look like. I just used unicode characters instead of proper icons to see what it looks like but I think it conveys the idea:
 
Screenshot 2024-09-20 at 12 03 32 PM

abaevbog:
It's really quite remarkable how many different ways there are to position two buttons!

yexingsha:
I like the version where the buttons are at the bottom of the saving section, with the items split off into their own progress section. I think it can work very well if we distinguish the two sections more:

middle-collapsed

middle-expanded

The version where the buttons are icons at the end of the first line can be somewhat confusing, since people will expect the x button at the top right corner to function as "Close" rather than "Cancel", and though the "Done" button should logically be placed next to "Cancel", it being at the top is pretty unusual. Also, "Done" is usually placed to the right of "Cancel", which means when the dialog is expanded, the "Cancel" button will shift to the left. A version with the hover interaction can look like this:


https://github.com/user-attachments/assets/9a5ed242-ac46-4dc3-90e3-31de28a83fba

adomasven:
I quite like the version with the expanding Cancel and Done buttons. @dstillman do you want to weigh in?

abaevbog:
For reference, I imagine it would look somewhat similar to this? I think we are running into a small issue of new buttons not looking very fitting into the existing design, and I feel like it's best to not implement the entire redesign of the connector in this PR. So I only changed the buttons in the header to look somewhat more uniform with lighter background and new svgs as icons. Otherwise, the chevron and `<select>` dropdown would look way too different from `X` and the checkmark.

https://github.com/user-attachments/assets/76a02a98-aac3-4cfb-be31-f08b842fa125

In this recording, I still have the progress items area as a scrollable container left over from the previous iteration. On one hand, it feels somewhat nicer to me than having the entire dialog overflow, on the other hand, functionally, it does not add anything if the buttons are at the top. So this part can be reverted, right? I ended up having a lot of changes while trying out different approaches, in case we want to revisit other options before committing.

yexingsha:
I think this is really good! The buttons fit in nicely with the current UI, and the scrollable progress item area works well. I agree that we can update the overall styling in a later PR.

adomasven:
Yeah, this is good. In the video the Done button seems to be bigger height than the other buttons and there's too much empty space where the Cancel button expands, so let's fix that and go with it.

abaevbog:
Yeah, the border of the button was of the wrong color and I did make the dialog a bit too wide... There were a few other smaller things with spacing that I tweaked. How is it looking now?

https://github.com/user-attachments/assets/a265001d-b6cb-4ff1-b958-22481a68b491

And I just cleaned up earlier unrelated changes and pushed these changes. One question: I added new `svg` files to use for done, cancel and the disclosure button - it does look like in the build process, there is a [step](https://github.com/zotero/zotero-connectors/blob/ea7878807658fa73dbf82976ce00c8431129c583/build.sh#L117) to fetch relevant images from the main repo but the submodule points at a pre-redesign commit that doesn't have all the new `svg` files, so I assumed just adding new files like this is they way to go for now?

Also, I was doing all the testing on firefox, so let me make sure everything looks right on chrome and safari.

abaevbog:
Chrome looks good to me now as well. Per https://github.com/zotero/safari-app-extension, do I need to be a part of the team to build and sign the safari extension? Sorry if that's an obvious question, it's been a while since I had to work with xcode.

dstillman:
Don't bother trying to test on Safari for now — it's too finicky. We'll check and report back.

adomasven:
> One question: I added new `svg` files to use for done, cancel and the disclosure button - it does look like in the build process, there is a [step](https://github.com/zotero/zotero-connectors/blob/ea7878807658fa73dbf82976ce00c8431129c583/build.sh#L117) to fetch relevant images from the main repo but the submodule points at a pre-redesign commit that doesn't have all the new `svg` files, so I assumed just adding new files like this is they way to go for now?

That's fine. There's a PR pending with updated icons (with the zotero submodule updated) that I am waiting an OK from Dan, but it shouldn't hold this up.

abaevbog:
Sounds good! Meanwhile, I added some final touches to remove unwanted spacing after buttons and to not show the "done" button unless target selector is expanded per https://github.com/zotero/zotero-connectors/pull/512#issuecomment-2364926136. I unintentionally reverted that change before pushing yesterday. With these final tweaks, I think I can un-draft this.

adomasven:
@abaevbog You should test the `manifestv3` build on chrome. I'll remove the `chrome` build as it's now fully phased out.

src/common/progressWindow/progressWindow.css
```diff
@@ -23,7 +16,18 @@ button:active {
 	padding: 6px 10px 6px;
 	min-height: 40px;
 	overflow: hidden;
-	position: relative
+	position: relative;
+	
+	/* Make progress box scrollable if there are too many items */
+	display: flex;
+	flex-direction: column;
+	overflow-y: scroll;
```
adomasven, line 24: This adds a permanent scrollbar on non-macOS. Should be `overflow-y: auto`

adomasven:
@abaevbog One more thing. We should enable this with a Zotero feature check. The response to /ping on Zotero should include `supportsSaveCancelling` and Connector should only show the cancel button in that case. Also no cancel button if saving to server.

abaevbog:
> You should test the manifestv3 build on chrome.

Gotcha, so from now on, I'll use `manifestv3` build folder for chrome but still `firefox` for firefox.

> One more thing. We should enable this with a Zotero feature check. 

Oh, that is a great point! Otherwise, one could have an updated connector that would try to hit `/connector/cancel` that the client might not yet handle. 

I pushed to the main repo PR to return `supportsSaveCancelling: true` together with other prefs when `/connector/ping` is hit. And here, we just check if it was received and hide the button if it wasn't OR if we're saving to web library.

src/common/ui/ProgressWindow.jsx
```diff
@@ -127,7 +127,7 @@ Zotero.UI.ProgressWindow = class ProgressWindow extends React.PureComponent {
 
 		// Check if the X button to cancel a saving process should be displayed
 		(async () => {
-			let notSavingToWebLibrary = await Zotero.Connector.getClientVersion();
+			let notSavingToWebLibrary = await Zotero.Connector.checkIsOnline();
```
adomasven, line 130: This issues a superfluous `/ping` call to the client. The ProgressWindow should already know whether it's saving to Zotero or the server.

abaevbog: It looks like we always try to [get selected collections](https://github.com/zotero/zotero-connectors/blob/ea7878807658fa73dbf82976ce00c8431129c583/src/common/inject/progressWindow_inject.js#L105) from the client (even if there is no zotero client running) and pass the targets (which are empty in case of saving to web) to `changeHeadline` which updates the state of ProgressWindow. And then, `this.state.targets` is used to [determine if the headline select](https://github.com/zotero/zotero-connectors/blob/ea7878807658fa73dbf82976ce00c8431129c583/src/common/ui/ProgressWindow.jsx#L595) for collections should be rendered. Is`this.state.targets` a reasonable way to determine if we are saving to desktop vs web? It does seems to work for instances that I tried, I just wasn't sure if there are times when it would not be the case.

adomasven: I think it's just `this.state.targetSelectorShown`? But target selector is written by Dan.

abaevbog: `targetSelectorShown` is [updated](https://github.com/zotero/zotero-connectors/blob/ea7878807658fa73dbf82976ce00c8431129c583/src/common/ui/ProgressWindow.jsx#L444) each time when the chevron is clicked, so I think it is about whether the target selector table and tags region is expanded or collapsed...

But judging by [this comment as well](https://github.com/zotero/zotero-connectors/blob/ea7878807658fa73dbf82976ce00c8431129c583/src/common/ui/ProgressWindow.jsx#L198), does it seem like the existence of `targets` should imply that we fetched them from the client, while their non-existence should imply that we'll save to the server? I feel like it does... Combined with the fact that in `updateFromClient`, "Saving to zotero.org" is set without any targets and "Saving to ..." is set with the targets, I am somewhat convinced that `this.state.targets` == saving to the client.

abaevbog: @dstillman, correct me if I am wrong on the above though!

abaevbog:
@adomasven, after the recent connector changes, we have a lot of conflicts here. Is it safe for me to look into rebasing this now? Or will there be more significant changes coming up and it's best to wait?

adomasven:
Yeah, sorry about that! No more major changes coming to saving for a while now.

adomasven:
@abaevbog This can be rebased whenever you're available.

abaevbog:
I didn't look much at the connector since when the refactoring to download attachments in the connector itself took place, and it looks much cleaner now! The logic is much easier to follow and, most importantly, it looks like the connector no longer has to keep checking in with Zotero desktop to see if the attachment is saved or not. I think now we can properly cancel the download [of an attachment in the connector](https://github.com/zotero/zotero-connectors/blob/9dd26f7bdda59d2d96e61d53289db544c7bd6fde/src/common/itemSaver_background.js#L183C19-L183C38), instead of waiting for everything to download and deleting items then.
So I think more of this cancelation logic will be moved to the connector, and here we'll mainly just handle deletion of items that were already saved.

src/common/http.js
```diff
@@ -127,6 +145,10 @@ Zotero.HTTP = new function() {
 				originalOptions.backoff++;
 				return Zotero.HTTP.request(method, url, originalOptions);
 			}
+			// Remove abort controller of a cancelable request that failed
+			if (options.cancellable) {
+				delete this.abortControllers[requestKey];
+			}
```
adomasven, line 151: Do this in a `finally` block.

src/common/itemSaver_background.js
```diff
@@ -222,6 +229,27 @@ Zotero.ItemSaver._unpackSafariAttachmentData = function(data) {
...
+	Zotero.HTTP.cancelPending();
+	// Wait a minute for all potential requests to get stopped and remove
+	// the session ID from the cancelled list
+	setTimeout(() => {
+		Zotero.debug(`Removing cancelled session ${sessionID}`);
+		Zotero.ItemSaver.cancelledSessions.delete(sessionID);
+	}, 1000 * 60);
+};
+
+Zotero.ItemSaver.assertNotCancelled = function (sessionID) {
```
adomasven, line 245: Needs docstring

adomasven:
> * Also, I see that Safari has some special handling with http requests. I am not sure if I can build the Safari extension, so let me know if there's anything else I should add for Safari.

It's not trivial to add it to Safari, and it's not strictly necessary, so let's skip it for now.

------

We now have a working testrunner for the Connector. We cannot run test code in Zotero content script environment on pages, but we can in the background service worker and in injected iframes (progress window). Would be good to add some tests here.

It might not be possible, or some points may not be testable due to inability to test code in es6 modules. A different test framework and additional build steps would be required for that and I haven't come around to doing it, but it would be great to have some tests if possible.

abaevbog:
@adomasven, while we're discussing the annotations dialog, I circled back here.

I did some minor refactoring and cleanup that should address the glitches you pointed out and some post-rebase bugs.

> If I save, cancel, then quickly press the button to save again, failed/cancelled attachment/snapshot saves from previous save appear in the progress window.

Right, thank you for pointing this out. This was caused by [the attachment progress handler](https://github.com/zotero/zotero-connectors/blob/4361593f853a12e313f56da980673acef1b939ec/src/common/inject/pageSaving.js#L181) always using the current session ID, even if it received an update about an attachment that was started with an old, cancelled sessionID. Now, sessionID is captured in a closure, so this should be taken care of. I also ran into a few smaller glitches of the popup window randomly reopening - we now make sure to not call `progressWindow.done` if the session is cancelled.


> Would be good to add some tests here.

The test runner is great! I added the tests for cancelling HTTP requests and to expect proper errors from `itemSaver` and `itemSaver_background`. There is a bit of logic happening in `pageSaving` too but it's closely tied to the translation process, so trying to add a test or two there, I ended up having so many `sinon.stub`s that it somewhat defeated the whole purpose of the test. All that really matters there is that we properly handle `session_cancel` errors to not let them propagate or call `progressWindow.done`. Let me know if there are any more aspects of this it's worth covering here.

> Aside from the scrolls appearing discussed in the other PR, the spacing looks weird now without the Cancel/Done buttons.

Finally, I am wondering if we should somewhat reconsider the "Done" and "Cancel" buttons... I agree, the spacing now is a bit weird when the popup is collapsed (with or without `supportsSaveCancelling`) due to the empty space to the right of the selector.

If the `X` button does not appear (no `supportsSaveCancelling`), we could let the selector take more space but we would still have to reserve space for the `Done` button when it appears if the popup is expanded. Otherwise, it would be unexpectedly jumping out when the popup expands, it would then receive `:hover` and expand more, which looks busy to me. So the twisty would still not properly pushed against the right edge.

https://github.com/user-attachments/assets/b37b3fd0-19f7-455c-b1aa-0e9c21424744


If `X` button appears (with `supportsSaveCancelling`), it gets even trickier. If we do not reserve extra space and have the selector expand all the way to the `X`, it would shrink the selector on hover, which looks strange. But most importantly, if you press on the twisty to expand the dialog, the mouse will find itself right over the "Cancel" button, and while trying this I found myself accidentally clicking on it a few times. Then, trying to move the mouse from expanded "Cancel" to the twisty, "Cancel" would shrink and move the twisty, forcing you to chase it with your mouse.

https://github.com/user-attachments/assets/56caa8c6-5afe-4bb0-962f-50a2c475bdb4

Moving the twisty would help with the issue above but I think it looks strange with the "Saving to" label right before it:

Screenshot 2025-07-11 at 2 56 03 PM

So with this, I am wondering if we can simplify some of this buttons-related logic a bit? Do the buttons need to collapse and expand? Maybe we can keep them as `X` and ` ✔`, and just add tooltips? And then, maybe they could be visible at all times to not deal with the "Done" button appearing and disappearing when the popup opens and closes?

Alternatively, I am wondering if the "Done" button is still necessary? I believe it was added so that one has something to click after typing in the tags. Now that the tags have autocomplete, if you add a tag, tag-bubbles appear above the tag input, which (to me) indicates that the tag has been saved.

adomasven:
> Alternatively, I am wondering if the "Done" button is still necessary? I believe it was added so that one has something to click after typing in the tags. Now that the tags have autocomplete, if you add a tag, tag-bubbles appear above the tag input, which (to me) indicates that the tag has been saved.

That change is not available to 7.0 users yet, but in the future it definitely wouldn't be necessary, I think we can remove it in this PR.

> So with this, I am wondering if we can simplify some of this buttons-related logic a bit? Do the buttons need to collapse and expand? Maybe we can keep them as `X` and `✔`, and just add tooltips?

I think the issue with this, especially if we remove the Done button is that the "X" just looks like a close button, not cancel. Maybe if we changed it to a trashcan it would make more sense. Otherwise I propose we make it expand, but it should overlap the target selector and the chevron upon expansion, instead of moving those elements. It might look bad/noisy, but it would make sense interaction-wise.

> And then, maybe they could be visible at all times.

That sounds good.

dstillman:
Done wasn't added just for tags, and I don't think it should be removed, at least without some sort of replacement. It's there so people have something to click on to 1) dismiss the dialog and 2) feel that they've saved any changes they've made. That obviously causes some people to think that they *need* to click Done, but that's better than an unsettling feeling that they're going to lose work or thinking that they have to wait several seconds for the panel to disappear.

This panel is largely modeled after Firefox's bookmark panel, and works the same way — the bookmark and any changes (folder selection, tags) are saved immediately, but there's still a Save button (in addition to Cancel).

dstillman:
(Looping @yexingsha back in for this.)

abaevbog:
I pushed what that suggestion might look like right now. Here, we have plenty of space for the buttons, so I think there would be no need to expand them on hover or hide them when the popup is collapsed. How does this look?

Screenshot 2025-07-14 at 11 05 56 AM

Screenshot 2025-07-14 at 11 06 02 AM

yexingsha:
I backread our discussion and it seems like we didn't really explore this option:

single-bottom-collapsed
multiple-bottom-collapsed

multiple-bottom-expanded

The objection at the time was that when saving multiple items, you would be chasing the Cancel button. But if we always set the height of the popup to max when saving multiple items, that shouldn't be an issue. I still like this option best.

If we put the buttons in the middle, this was my suggestion for styling:

middle-expanded
middle-collapsed

abaevbog:
Oh yeah, that's right! We touched upon it but when the potential of chasing the "Cancel" button moved by the expanding list of progress item was brought up, we moved on to other ideas.

I think both options are equally valid, each with its trade offs. The buttons above the progress items list is odd because "Done" and "Cancel" anywhere that's not the bottom of the popup is unusual, no matter how we style it. Having buttons at the bottom while reserving a predefined height for the progress items can be odd because there would be empty unused space in that section if there are not too enough items being saved.

I feel like which option is better depends on which use case is more common. If users frequently save multiple items, having the reserved space seems better to me, since it will eventually fill out. But if users mainly save just one item, that reserved space would end up being wasted most of the time, so adding items at the bottom may be better to not have to guess how much space to reserve. Does that seem reasonable? @adomasven, what do you think is a more frequent use case?

While working on this initially, I didn't touch the exiting UI layout, with all the text inputs, backgrounds, etc. remaining as is. But I suppose now that we are moving the buttons back into the popup, the contrast between new "Done" and "Cancel" buttons and the old UI is quite apparent. So I think I'll update some of the stylesheet meanwhile.

yexingsha:
We always know how many items are being saved before the popup appears, right? So we don't need to reserve the space if the user is saving only one item. Or maybe we can always calculate how much space to reserve, using [number of items] * [maximum item height of 4(parent, pdf, snapshot, note?) lines].

abaevbog:
> We always know how many items are being saved before the popup appears, right?

We do know how many top-level items are being saved before the popup appears:

https://github.com/zotero/zotero-connectors/blob/4361593f853a12e313f56da980673acef1b939ec/src/common/inject/pageSaving.js#L224-L228

But at this point, we don't know what notes/attachments those top-level items would have. After the translation runs on those individual items, then we get that info. This process can take a few seconds, depending on how many items are being saved. For instance, if I save 10 items from JSTOR, it takes 2-3 seconds before we know that each item has only 1 attachment:

https://github.com/user-attachments/assets/56d52a22-087a-4112-b8f1-20749db9b281

So if we reserve the height of an attachment + note + snapshot for each item initially, after 2-3 seconds we would have to shrink it back. In this example, I am saving a lot of items so it probably would work smoother with fewer items, and this delay would be smaller.

There can be instances when a translation would get stuck and throw an error after a delay (e.g. I ran into some networking issues a few times). Then, the whole process takes a bit longer. That actually brings up another question - if errors are thrown, we have messages that need to be displayed. If we try to reserve space for item rows, how would the errors appear?Currently, they pop up below the list of items, like this:

Screenshot 2025-07-14 at 2 49 22 PM

abaevbog:
I moved the "Cancel" button back to the top-right corner of the dialog, together with some updates to the UI so it looks closer to the design mockups.

This is what it's looking like now with the old `X` icon:

Screenshot 2025-07-15 at 2 07 32 PM

The "Done" icon looks quite awkward, doesn't it? Just as an experiment, I moved it back to its original spot after the tags input. It doesn't make sense for it to be there but I feel like it looks better vs just having all that empty space at the bottom:

Screenshot 2025-07-15 at 2 08 10 PM

abaevbog:
I quite like this! I think the trash icon works very well here.

This is the updated popup. How does this look?

Screenshot 2025-07-15 at 3 37 10 PM
Screenshot 2025-07-15 at 3 37 05 PM

With this, the UI is updated partially. The two things remaining would be to support dark mode as well as to use updated svg item icons in the progress list. The item type icons also appear in the browser toolbar (the main Zotero connector icon). So would these two things have to happen at the same time? If so, perhaps it's best to just use old icons for now. But otherwise, @adomasven, I think we would need to update the `src/zotero` submodule to point at a more recent, post-redesign zotero commit that would have the new svg icons. Is there any reason to not do this at the moment?

yexingsha:
I think they look pretty good! Though it seems like we aren't using macOS native UI components in the connector, and there is no way to do that, right? If that's the case then I will switch them to our own style so that they look better on different platforms.

dstillman:
This is nice!

My concern, though, is that it means having a fairly prominent trash icon on one of our most iconic UI components, which I don't love. I also think it's a bit premature to add a cancel button to the collapsed panel before we've added the Z badges, which may greatly reduce the frequency of accidental clicks.

So what if we just move the small trash button down to the left of the Done button? Then there's still an easy way to delete the item without switching to Zotero, but we're not sullying the collapsed popup. And then once this exists in the expanded popup and we've added the Z badges, we can see if we're inundated with complaints from people who want to cancel without expanding the popup. But I'm hoping this would be sufficient.

abaevbog:
Oh, @yexingsha, one small thing I noticed. The trash icon looks slightly closer to the right edge of the button than the left at the moment. That seems to be the case in the 16x16 icon I fetched from figma. Am I looking at the wrong thing? Or could we make it slightly wider/narrower to have the same space on each side? That's probably cleaner than shifting it by one pixel with css

Screenshot 2025-07-16 at 12 25 37 AM

abaevbog:
> Set the corner radius of the collection list box to 0px

Ah, yes. I didn't notice that it's the only component with no border-radius, sorry about that!

> "Tags", trashcan icon and "Done" should be vertically aligned to the center

They _should_ be aligned css-wise - but I guess with text it looks slightly off. I pushed it up by a few pixels, does it look better now? 

> Set the focus ring of the input box to --color-focus-border (or accent-blue50 if that's not available), and get rid of the white outer border

Good catch! I didn't even notice there was a weird white border around the outline. I changed it to be `SelectedItem` (which seems to be what `color-focus-border` is based on in the majority of cases)  or `--accent-blue50` if it doesn't exist.

> Move the autocomplete list down so that it's 4px away from the input box, and doesn't obstruct the focus ring

Good point!


Screenshot 2025-07-16 at 1 19 00 PM

abaevbog:
> Just need a little more padding at the very bottom to match the top padding of the section. Also, the <select> shouldn't shrink when a different collection is selected

Another good catch! The padding on the popup was all over the place, I set it for 12px as in the mockup. And `<select>` will now expand to the end, regardless of the selected collection.

adomasven:
> If so, perhaps it's best to just use old icons for now. But otherwise, @adomasven, I think we would need to update the `src/zotero` submodule to point at a more recent, post-redesign zotero commit that would have the new svg icons. Is there any reason to not do this at the moment?

We're getting the current item tree and collection icons from the zotero commit to which this repo points, so we cannot just trivially update it. There's a [PR pending for new icons](https://github.com/zotero/zotero-connectors/pull/502) and update to zotero submodule. @dstillman needs to review and merge it.

abaevbog:
OK, gotcha. Thanks for the explanation.

I have 4 svg icons added here - chevron, trash (for cancel button), collection, and library (for the target selector). In all cases, I inserted them inline to be able to set the `fill` color of the icons. I see in the linked PR, I could have actually used `maskImage` instead. In any case, there would be some conflict between the current changes here and the PR with new icons. 

It might make sense to wait for icons PR https://github.com/zotero/zotero-connectors/pull/502 to be merged first, so that I could rebase on master with new icons, fetch `collection` and `library` icons from the location that they are supposed to be at, and make sure that everything works well with the dark mode here.

PR #516, opened 2024-10-25
ProgressWindow: Fix error when adding line containing `'` to a11y log

AbeJellinek:
Or anything else that violates CSS string syntax.


adomasven:
Where is this causing an issue?

PR #517, opened 2024-10-30
Update testrunner for Firefox

maxsu:
This patch gets tests running on firefox 131. We install the addon in temporary mode and drop the preference hack.

### Issue
Firefox test run fails with

```
WebDriverError: Could not install add-on
```

### Changes
1. Install addon in temporary mode
- Tested in Firefox 131 =>  fixes issue
2. Remove broken `xpinstall.signatures.required` code
- Tested in Firefox 131 => fix remains stable

### Docs
- [Selenium Docs: Add-ons: Unsigned installation](https://www.selenium.dev/documentation/webdriver/browsers/firefox/#add-ons)

### DIFF
--- a/test/seleniumTests.js
+++ b/test/seleniumTests.js
@@ -137,14 +137,7 @@ if ('TEST_FX' in process.env) {
 				.setFirefoxOptions(options)
 				.build();
 
-			await driver.setContext(firefox.Context.CHROME);
-			await driver.executeScript(`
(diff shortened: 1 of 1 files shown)


adomasven:
We're not really using this testrunner anymore since running evals has been severely restricted with MV3 (on which the old tests rely heavily) and tests are likely to fail. However since the development of these tests new selenium and browser APIs have been added that would allow to write proper tests automating the browser and testing the plugin for behavior.

Issue #310, opened 2020-04-20
Tag autocomplete in save popup


dstillman:
https://forums.zotero.org/discussion/82644/display-existing-tags-when-saving-from-the-web and I think some other threads

ashwinvis:
Seems like a duplicate of #206.

Having the tags input box visible while clicking "Save to Zotero" will also be nice to have ... instead of having to press the drop down arrow everytime as it is now.

dstillman:
https://forums.zotero.org/discussion/comment/415098/#Comment_415098

dstillman:
https://forums.zotero.org/discussion/110482/feature-request-drop-down-list-of-existing-tags-in-web-clipper-a-la-evernote

manbackintown:
This feature is also requested here zotero/zotero#1706

PR #529, opened 2025-02-11
add autocompletion to tags input

abaevbog:
- if the desktop zotero app passes a list of tags, render them as datalist options for the tags input
- the user adds a newly typed tag (matching one from autocomplete or not) via Enter. Those tags will be displayed above the input as bubbles with an X button to remove them
- if tags input contains text matching exactly one tag, it will be added automatically as well. Due to browser handling the actual autocompletion, we cannot add a listener to add the tag when autocomplete popup entry is selected. But when there are no other canditates, this can save the user an extra Enter keypress
- if the target collection is switched to one from another library, selected tags are cleared, since the new library may have completely different tags
- one can navigate between tags' "remove" buttons with left/right arrows
- when the update is sent to the zotero desktop app, join selected tags with a `,` to preserve current format ignoring everything else left in the input
- if the desktop app does not send a list of tag, and there are no selected tags, fallback to sending text in the input as tags. This is for backwards compatibility if one has an older version of the desktop app.

Fixes: #310

https://github.com/user-attachments/assets/8ea924e9-3510-4cdf-865d-cab84a655c87


abaevbog:
I tested this on Firefox and Chrome with and without the linked PR that passes sends the tags. Not sure what are the steps for building on Safari though

src/common/ui/ProgressWindow.jsx
```diff
@@ -202,6 +208,14 @@ Zotero.UI.ProgressWindow = class ProgressWindow extends React.PureComponent {
...
 		let alert = this.done ? Zotero.getString("progressWindow_alreadySaved") : `${text} ${target.name}`;
 		document.getElementById("messageAlert").textContent = alert
+		// Tags array contains objects {tag : ""} that may contain duplicate values due to different types.
+		// Unwrap the tag objects and deduplicate tags values.
+		// this.existingTags will be an object of the form {libraryID: [tag1, tag2, ...]}.
+		let dedupedTags = {};
+		Object.entries(tags || []).forEach(([libraryID, tagArr]) => {
+			dedupedTags[libraryID] = [...new Set(tagArr.map(item => item.tag))];
+		});
+		this.existingTags = dedupedTags;
```
adomasven, line 218: Seems like this would be more appropriate to do in `updateFromClient()` which does some preprocessing of data for the Progress Window.

adomasven:
Excellent work! The auto-confirm of tags when typing manually is a bit weird and i imagine could cause people some issues if you want to add a new tag that has an old tag as a prefix. E.g. p1 and p10 for priorities or whatever. Can we not detect it from mouse events, or maybe `blur` or something is called on click?

Using my test library with a bunch of tags I see this

image

Can we limit the height of the autocomplete?

abaevbog:
> Can we not detect it from mouse events, or maybe blur or something is called on click?

That is the main limitation of using `input` with `<datalist>`. The autocomplete popup is handled by the browser, so click or keydown events don't fire on the popups. The only way to find out what has been clicked seems to be by checking input's value against existing `<options>`s. This bit is not explicitly documented (or maybe just not in [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist), where I normally check for component properties)  but the first search result I get is this [stackoverflow question](https://stackoverflow.com/questions/30022728/perform-action-when-clicking-html5-datalist-option) that also talks about this and other hacky workarounds.

> Can we limit the height of the autocomplete?

Same issue here due to the popup being handled by the browser. In firefox, it is a fairly small autocomplete field with just 3-4 visible options at a time. In chrome, it is a larger dropdown.

When briefly asking  dstillman about this issue, `<datalist>` was mentioned as a candidate to not have to worry about autocomplete. It looks like it may be too limiting for us though... Mozilla docs also mention a few screen reader/browser combinations where autocomplete popup may not work well. So I am now thinking maybe it actually is worth implementing a minimal version of autocomplete popup that we can style ourselves and listen to all sorts of events.

dstillman:
If `datalist` is too limiting, let’s look for a small existing library with no dependencies. No need to write the millionth version of this ourselves.

dstillman:
We’ll want some styling from @yexingsha for the tags.

abaevbog:
The `datalist` does seem a bit too restrictive ... Even not being able to properly tell if the user clicked on an autocomplete suggestion is inconvenient. And we are at the mercy of the browser to style it.

I actually might be almost done with putting together a standalone tags autocomplete component. It felt a bit silly to be implementing it for the 1000th time but it was not that bad. And less dependencies, which is a plus. I can push it after a minor cleanup, and we can see if there is anything significant missing there before looking for 3rd party library? Or do you think it's not really worth it, and it's better to just pick a library right away?

This is what it's looking like right now:

https://github.com/user-attachments/assets/eccac77f-2384-467e-9585-7277813037c3

dstillman:
No, if you have something already, that’s fine.

abaevbog:
> We’ll want some styling from @yexingsha for the tags (and maybe the menu itself, though I guess it depends what we get for free from a third-party library).

Oh and yeah, any and all styling suggestions for the popup and the selected tags bubbles are very appreciated!

src/common/ui/ProgressWindow.jsx
abaevbog: That makes a lot of sense, yeah. I completely re-did and re-pushed the last commit with the new `<TagsInput>` component that has our own autocomplete popup. I moved this bit of pre-processing to `updateFromClient` from here.

dstillman:
> The best solution that I could come up with was to add extra padding below the tags input area, which stretches progress window and forces iframe to resize. I think it looks fairly natural and better than placing the autocomplete popup above the input.

That's clever — totally wrong and nonsensical, but you don't really notice because the popup covers the space! Except, is that the case if there's a single match? From your recording, it looks like it's pushed down enough that two rows covers the empty space, but what if there's only one row?

adomasven:
@abaevbog You can make iframes transparent, can't you? That would still maybe make them clickable in a way that we don't want, but this doesn't have to be such a problem.

dstillman:
I assume it's the styled div being expanded that's relevant, not the iframe, which is already transparent. But yeah, that should be solvable without the padding hack…

abaevbog:
> That's clever — totally wrong and nonsensical, but you don't really notice because the popup covers the space! Except, is that the case if there's a single match? From your recording, it looks like it's pushed down enough that two rows covers the empty space, but what if there's only one row?

It was a very "creative" approach which, unfortunately, works way worse in circumstances when there are more than 2 rows. If I save multiple items, then the iframe is actually quite tall, so there should not be any extra space added for the popup. Adding padding in that case is not quite right.

So i reworked that logic to essentially check if the popup fits within the iframe, and if not - add extra height that the autocomplete popup needs while resizing the iframe in the main `ProgressWindow` component. I also tweaked the stylesheet a bit so that the popup is positioned relative to the iframe, and not the `.ProgressWindow-box`. With this, I think we get the behavior that we want.

If adding multiple items, autocomplete popup does not change anything because there is plenty of space for it:

https://github.com/user-attachments/assets/5eaecd5b-2018-4ed1-bf4a-f721dc2f4f97


If adding a single item, `<TagsInput>` will ask the `<ProgressWindow>` to resize the iframe so that it is tall enough to fully display the popup. (I haven't been able to find a good url with just metadata, without attachment, to see what would happen if only one row is being saved but the same logic should apply.) 


https://github.com/user-attachments/assets/be38cd2b-3e1e-4637-83dc-fad2e14a53bd


@yexingsha I added the stylesheet tweaks you pointed out earlier. How does it look now? I wonder if 13px is too much for the autocomplete menu in the recordings above? Everything else (e.g. "Done" button next to it) is 11px. Do the menu options stand out a bit too much? If not, and it's just my personal take (which is often wrong), should we make the menu a bit taller? Currently, it's 80px.

yexingsha:
Looks like the current dialog has a mishmash of font sizes: "Saving to" is 13px, the collection tree is 13px, all input boxes are 12px, the items and attachments to be saved are 12px, and the "Done" button is 11px. Ideally everything aside from tags should be 13px, and we should definitely fix the weirdly small "Done" button.

The issue with the menu is the height of each item, which should be reduced to 20px. Also, it seems that its text is not vertically centered?

Other than that, the `x` button in tags should be colored `fill-secondary`, and have less space before and after. In the mocks the `x` button is 12px, and the tags have `padding: 1px 3px 1px 4px; gap: 2px;`

yexingsha:
Looks good to me!

adomasven:
@dstillman Unless you have further comments I think we can merge this.

adomasven:
The cursor should change to pointer when on the "x". I know we don't do this anywhere else in this dialog, but the highlight when you hover over X is too minimal.

It's somewhat weird with tab navigation that you can highlight the tags (not Xes on them) and pressing Space removes those tags. I know it's the only meaningful interaction, but it should probably show the outline on the X button.

abaevbog:
> I also have tags in Zotero that contain , (comma) symbols. When trying to auto-select them, they get split into two separate tags.

Oooh, yeah, that's a bit of an issue. My idea here was to plug this update into existing workflow as seamlessly as possible, so tags are joined with a `,` before being sent. I guess it does do exactly that because, at the moment, typing a tag that has a comma would add two tags as well. In any case, I think we should then send tags as a proper array and just be ready to handle that, as well as the old, comma-separated format in desktop.

abaevbog:
With the last few updates, the connector will send tags as an array, so tags that have actual commas will not get lost. I updated the linked PR in the main repo as well to handle tags passed as an array. If tags are sent as a string (e.g. from an older connector), then we'll split them on `,` the same way it happens now. I also included a new flag to indicate if tag autocomplete is supported to be able to reliably decide in which format the connector should send the tags.

I also included the minor style fixes to place outline around the `X` and add a pointer on `X` as well. 

> I've rebased this since the merge of note PR, but seeing

I am struggling with reproducing this, however. I don't see it on any of the platforms/browsers that I tried. Where are you seeing this?

adomasven:
> I am struggling with reproducing this, however. I don't see it on any of the platforms/browsers that I tried. Where are you seeing this?

Occurs on Chrome on Ubuntu, although the scroll disappears at 140% zoom (but not other levels). It's due to the font size change, and the line-height not being enough. It wouldn't hurt to rework the layout to use flex, otherwise this requires changing line-height, but it's kind of arbitrary.

Dadoopant:
@abaevbog  

Hello All. I'd love to be able to to configure my Zotero to do this. I am a novice. Do I need to change some coating in Zotero to do this? I'm unable to find instructions on how to do that. I am running Zoteroi 7.0.17 and have just updated the chrome connector, but I'm unable to get auto complete or a list of tags for any collections. Advice for a ranked novice would be most welcome.

I'm unclear if this is a set of instructions I need to complete to make this work. 
if the desktop zotero app passes a list of tags, render them as datalist options for the tags input
the user adds a newly typed tag (matching one from autocomplete or not) via Enter. Those tags will be displayed above the input as bubbles with an X button to remove them
if tags input contains text matching exactly one tag, it will be added automatically as well. Due to browser handling the actual autocompletion, we cannot add a listener to add the tag when autocomplete popup entry is selected. But when there are no other canditates, this can save the user an extra Enter keypress
if the target collection is switched to one from another library, selected tags are cleared, since the new library may have completely different tags
one can navigate between tags' "remove" buttons with left/right arrows
when the update is sent to the zotero desktop app, join selected tags with a , to preserve current format ignoring everything else left in the input
if the desktop app does not send a list of tag, and there are no selected tags, fallback to sending text in the input as tags. This is for backwards compatibility if one has an older version of the desktop app.


Thank you?

Dadoopant:
Thank you!

Okay. 

Any idea when it might happen?

Issue #376, opened 2022-02-15
Add note field to save popup


dstillman:
https://twitter.com/bianca_oli_per/status/1493488321691009031

Don't want to make the save popup too complicated, but a three-line text area that gets turned into a child note might be nice.

dstillman:
https://forums.zotero.org/discussion/91285/add-coment-from-the-zotero-connector-feature-request

dstillman:
https://forums.zotero.org/discussion/113704/zotero-connector-has-no-way-of-adding-a-note

PR #533, opened 2025-02-25
Add note field to progress window

abaevbog:
So one can add a child note to items that are being saved. 

- if client passes `canUserAddNote` flag in ping response, display a textarea note field above tag selector
- the window will remain open if the focus is in the textarea, same as with tags
- on blur, note content is passed to the client to save as a child note
- Enter from the note field will add a new line. Shift-Enter can close the dialog.

Needs: https://github.com/zotero/zotero/pull/5064
Fixes: #376

Brief demo: https://www.dropbox.com/scl/fi/k7jm6uc3t4syve2vs8z6a/connector_note_demo.mov?rlkey=3h91ujz6qzeyzyd72nsv50uyw&st=gyuac1x4

### DIFF
--- a/src/common/connector.js
+++ b/src/common/connector.js
@@ -72,6 +72,7 @@ Zotero.Connector = new function() {
 			Zotero.Connector.googleDocsAddNoteEnabled = !!response.prefs.googleDocsAddNoteEnabled;
 			Zotero.Connector.googleDocsCitationExplorerEnabled = !!response.prefs.googleDocsCitationExplorerEnabled;
 			Zotero.Connector.supportsAttachmentUpload = !!response.prefs.supportsAttachmentUpload;
+			Zotero.Connector.canUserAddNote = !!response.prefs.canUserAddNote;
 			if (response.prefs.translatorsHash) {
 				(async () => {
 					let sorted = !!response.prefs.sortedTranslatorHash;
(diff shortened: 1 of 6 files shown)


abaevbog:
The positioning of the `textarea` may need to be tweaked with a bit, especially with the tags autocomplete being added. We could only show the `textarea` if one presses a separate button, and keep it hidden otherwise - not sure.

adomasven:
This is good, but the textarea should either be more than 2 rows or auto-expand on focus/typing to at least 5 rows. Also your uploaded demo is inaccessible.

abaevbog:
> the textarea should either be more than 2 rows or auto-expand on focus/typing to at least 5 rows.

I figured that we don't want to make the `textarea` too big because the iframe is fairly small. I suppose not everyone would be regularly adding notes like this, so it's good to keep it as compact as possible. That said, 2 lines is indeed quite tight. First I tried to expand it to 5 lines on focus, same way we do it in `editable-text` with multiline fields in itembox. But it means that on blur, the field can potentially shrink, resize of the `iframe`, and potentially make one mis-click a component they were not aiming for. So maybe auto-expanding is the best solution here? Now, the field will expand to have the same height as the `scrollHeight` up to `6em`, which fits 5 lines.  On blur, it remains as is, without causing any resizing jumpiness.

https://github.com/user-attachments/assets/c7ecaf8a-51db-4068-a8df-ce19ee983731

> Also your uploaded demo is inaccessible.

Sorry about that! I must've not copied some path params in the url, which still worked fine for me cause I was logged in. I updated the link in the first comment.

adomasven:
Sorry @abaevbog can, we rebase this? Also, can we change the font of the textarea to be something better than courier new?

Issue #509, opened 2024-08-29
Save session stays open after failed PDF download


AbeJellinek:
Connector 5.0.144, Firefox 129.0.2
Example URL: https://link.springer.com/chapter/10.1007/978-3-319-23168-6_1

The PDF fails to download without institutional access, and no open-access PDFs are available. The connector continues making `/connector/sessionProgress` requests every two seconds for a couple minutes after the save progress window has disappeared. Zotero responds with `{ done: false }`.

```
zotero(5)(+0000001): HTTP/1.0 200 OK
X-Zotero-Version: 7.0.4.SOURCE.051d3384c
X-Zotero-Connector-API-Version: 2
Content-Type: application/json

{"items":[{"id":"oxO3jSh1","title":"Parkinson’s Disease","itemType":"bookSection","attachments":[{"id":"5n1QTrDl_1","title":"Full Text PDF","contentType":"application/pdf","progress":false}],"progress":100}],"done":false}
```

Issue #474, opened 2024-05-16
Fetch all attachments using xhr/fetch in the Connector and save directly to Zotero and Server


adomasven:
Seems like when the Connector was initially written there was no XHR support for Blob and ArrayBuffer types. Technically it's not very efficient to do this, but we are facing increasingly more complicated bot protection and it's unlikely to get better in the future.

dstillman:
(Inspired by https://forums.zotero.org/discussion/114431/pdf-will-not-save-to-zotero, among others)

dstillman:
https://forums.zotero.org/discussion/comment/465245/#Comment_465245

adomasven:
https://forums.zotero.org/discussion/116557/zotero-ezproxy-issue

adomasven:
While this works well for regular downloads, for some sites (ScienceDirect!) that use a JS redirect it will not without custom handling:

- Add a hidden iframe on the page with sandbox settings that disallows file downloads (so the user doesn't get prompted to save a file)
- Monitor for attempts to navigate to the expected mime-type (either navigation gets aborted after getting headers with content-disposition by iframe sandbox, or it succeeds, which means we needlessly load the file in the iframe)
- Then refetch the same page using XHR

Unfortunately at least for ScienceDirect on Safari this doesn't work without some additional custom handholding, because we do not have full cookie access there. We need to make the fetch request for the PDF from the sciencedirect page (not background) as it's allowed per their CSP policy, but this can change at any point.

The biggest drawback here is that we're not equipped to display a captcha window on the browser if anything goes wrong, and even if we opened a new captcha tab, we wouldn't be able to do much with it since we cannot grab the loaded file directly from the browser without using XHR. Also it might trigger a save file prompt if we did this.

Also if some of this breaks we will be hostage to the browser extension approval processes and cannot update as fast as we can with the client.

The other option is to continue using Zotero BrowserDownload for pages where we need JS redirect, but that defeats one of the more exciting parts of this change.

dstillman:
https://forums.zotero.org/discussion/120716/zotero-7-is-not-saving-pdfs-in-safari

PR #535, opened 2025-03-21
Fetch attachments in Connector and pass to Zotero on save

### LINKED ISSUE #414: Saving to server doesn't work on PDFs
https://forums.zotero.org/discussion/101314/zotero-connector-not-saving-pdfs

Not totally sure why we don't allow this, but I guess we should — as I say there, we do allow it on iOS, where PDF metadata retrieval also isn't available. Here it's a bit more likely to just be a mistake, but I'm not sure that's a reason not to support this one feature when we support other things.

adomasven:
Changes the Zotero Connector so that attachments (PDF, etc) are downloaded in the connector with a background fetch and passed directly to Zotero. The Zotero Connector is backwards compatible with older Zotero versions which do not support receiving attachments from the Connector (new version of Zotero will not be backwards compatible with old connectors).

Supports advanced captcha handling (under whitelist, currently only sciencedirect.com), where if a PDF fetch fails we will load the pdf in a hidden frame and intercept upon content-disposition: attachment and then refetch with xhr. If hidden iframe fails, we pop open a window for the user to solve the captcha, and once again intercept and refetch with xhr. Seems to work very well. This is not supported on Safari, and we will still want to keep the captcha handling code in Zotero too (for magic wand, etc).

Generally means that attachment saving when access is provided by a proxy will now work in pretty much every case, since we don't need to pass credentials to Zotero and pretend we are the user's browser there.

Also fixed attachment saving to server.

Safari also needs: https://github.com/zotero/safari-app-extension/pull/7
Corresponding Zotero PR: https://github.com/zotero/zotero/pull/5148

Closes https://github.com/zotero/zotero-connectors/issues/474 , closes #414, closes https://github.com/zotero/zotero-connectors/issues/509

### DIFF
--- a/src/common/inject/pageSaving.js
+++ b/src/common/inject/pageSaving.js
@@ -320,9 +320,10 @@ let PageSaving = {
 		var result = await Zotero.Inject.checkActionToServer();
 		if (!result) return;
 
+		const isOnline = await Zotero.Connector.checkIsOnline();
 		const supportsAttachmentUpload = await Zotero.Connector.getPref('supportsAttachmentUpload');
 
-		if (document.contentType !== 'text/html' && supportsAttachmentUpload) {
+		if (document.contentType !== 'text/html' && (supportsAttachmentUpload || !isOnline)) {
 			return await this._saveAsStandaloneAttachment({title, saveSnapshot});
 		}
 		return await this._saveAsWebpage({title, saveSnapshot});
@@ -410,31 +411,26 @@ let PageSaving = {
 		} catch (e) {
 			// Client unavailable
 			if (e.status === 0) {
-				// Attempt saving to server if not pdf
-				if (document.contentType != 'application/pdf') {
-					let itemSaver = new Zotero.Translate.ItemSaver({});
-					this.sessionDetails.itemSaver = itemSaver;
-					let items = await itemSaver.saveAsWebpage();
-					if (items.length) {
-						Zotero.Messaging.sendMessage(
-							"progressWindow.itemProgress",
-							{
-								id: title,
-								iconSrc: Zotero.ItemTypes.getImageSrc(image),
-								title,
-								parentItem: false,
-								progress: 100
-							}
-						);
-					}
-					Zotero.Messaging.sendMessage("progressWindow.done", [true]);
-					return;
-				} else {
-					Zotero.Messaging.sendMessage("progressWindow.done", [false, 'clientRequired']);
+				let itemSaver = new Zotero.Translate.ItemSaver({});
+				this.sessionDetails.itemSaver = itemSaver;
+				let items = await itemSaver.saveAsWebpage();
+				if (items.length) {
+					Zotero.Messaging.sendMessage(
+						"progressWindow.itemProgress",
+						{
+							id: title,
+							iconSrc: Zotero.ItemTypes.getImageSrc(image),
+							title,
+							parentItem: false,
+							progress: 100
+						}
+					);
 				}
+				Zotero.Messaging.sendMessage("progressWindow.done", [true]);
+				return;
 			}
-				// Unexpected error, including a timeout (which we don't want to
-				// result in a save to the server, because it's possible the request
+			// Unexpected error, including a timeout (which we don't want to
+			// result in a save to the server, because it's possible the request
 			// will still be processed)
 			else if (!e.value || e.value.libraryEditable != false) {
 				Zotero.Messaging.sendMessage("progressWindow.done", [false, 'unexpectedError']);
@@ -476,7 +472,8 @@ let PageSaving = {
 		let standaloneAttachment = {
 			url: document.location.toString(),
 			mimeType: document.contentType,
-			title
+			title,
+			linkMode: "imported_url"
 		}
 
 		try {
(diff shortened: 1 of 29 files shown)


adomasven:
This PR needs to be merged and deployed first (along with Safari changes). The Zotero PR shouldn't be deployed until this is fully rolled out for Connector users as the changes in the Zotero client are not backwards-compatible with earlier versions of Connector.

dstillman:
Not sure how much of the save-to-server logic has actually changed, or what we actually supported before, but if I'm at my storage quota, I get a misleading error message that suggests the whole save has failed:

Screenshot 2025-03-28 at 5 03 12 AM

And I get two broken attachments, since the attachment items are created but the file requests fail with 413s, and the attachment items aren't deleted after. I can open a separate issue for this if this isn't new.

dstillman:
But the proper behavior here would be red X for the attachments and automatic DELETE on the failed attachment items, plus ideally a message below about the quota.

adomasven:
> And I get two broken attachments, since the attachment items are created but the file requests fail with 413s, and the attachment items aren't deleted after. I can open a separate issue for this if this isn't new.

Since attachment upload didn't used to work before (everything got attached as links), this is new, but probably doesn't have to hold this PR up.

src/common/itemSaver.js
```diff
@@ -127,64 +124,228 @@ ItemSaver.prototype = {
...
+	},
+	
+	async saveOAAttachment(item, attachmentCallback) {
+		let attachment = item.attachments.find(a => a.isPrimary);
+		try {
+			// Check if we can get an OA PDF from Zotero
+			if (typeof item.hasAlternativeAttachment === "undefined") {
+				// Change line for the primary attachment if we're going to look for OA alternatives
+				if (attachment) {
+					attachment.title = "Searching for Open Access files…";
```
dstillman, line 301: We'll need to localize these.

src/common/inject/pageSaving.js
```diff
@@ -627,6 +598,46 @@ let PageSaving = {
 		
 		var sessionID = this._initSession(translatorID, options);
 		return await this.saveAsWebpage({sessionID, title, snapshot: options.snapshot, resave: options.resave});
+	},
+
+	/**
+	 * Updates the session with the given data.
+	 * @param {Object} data - The data to update the session with.
+	 * @param {String} data.targetId - The target ID
+	 * @param {Boolean} data.resaveFiles - Whether files should be resaved
```
dstillman, line 616: resaveAttachments?

src/common/inject/progressWindow_inject.js
```diff
@@ -56,6 +56,7 @@ if (isTopWindow) {
 	var nextSessionUpdateData;
 	
 	var isReadOnly = false;
+	var isFiledEditableOnly = false;
```
dstillman, line 59: isFilesEditable?

src/common/itemSaver.js
```diff
@@ -127,64 +124,228 @@ ItemSaver.prototype = {
...
+					itemID: item.id
+				});
+			}
+			if (!item.hasAttachmentResolvers) {
+				if (attachment) {
+					attachment.title = Zotero.getString("progressWindow_OA_notFound");
+					attachmentCallback(attachment, false)
+				}
+				return;
+			}
```
dstillman, line 315: I don't think the logic here is right (and I think is just like this because of the original incorrect `HasOAAttachments` endpoint name).

1) If there are no resolvers, we don't need to say anything about OA versions at all — it should just stay on the original title (presumably with the red X).

2) We already show "Open Access PDF" when we're trying the actual download. `hasAttachmentResolvers` should be very quick. 

So I don't think we need the `searching` or `notFound` strings at all?

adomasven: Right, if there, but it should be mimicking what we did before:
1. If the original translator didn't specify a PDF attachment, we don't display anything unless we find an OA PDF
2. If the original translator does specify a PDF attachment and it fails to load, we replace it with "Searching for Open Access PDFs" provided there are resolvers.
3. If we fail to get an OA PDF, then we show the `failedToGetPDF` string.

Arguably, for 2 maybe we should not be replacing the line for the attachment that failed to download via the browser, and just append a line about search for OA PDFs. And yes, we don't need the `notFound` string.

dstillman:
Regression: Saving a PDF directly shows a "Snapshot" entry below the attachment.

Screenshot 2025-04-02 at 1 43 18 AM

dstillman:
Should saving a PDF directly to the server not work?

Screenshot 2025-04-02 at 1 48 02 AM

dstillman:
The save popup isn't closing automatically when saving to server from https://arxiv.org/abs/2504.00054:

Screenshot 2025-04-02 at 1 50 16 AM

In console:

> Uncaught (in promise) TypeError: Message length exceeded maximum allowed length.  background-worker.js:1 
> Promise.then |   |  
>  | onMessage | @ | browser-polyfill.js:887

adomasven:
> Should saving a PDF directly to the server not work?

It didn't historically work, but we can definitely add this.

adomasven:
Addressed and also some additional fixes for saving to server

dstillman:
Same snapshot failure on any translator page, I think, including https://arxiv.org/abs/2504.00054 above, which was working before.

Screenshot 2025-04-02 at 4 42 12 AM

And there's no snapshot attempt on https://www.zotero.org/, which doesn't have a translator, even though I have the pref enabled.

dstillman:
And an error directly saving the above PDF to server:

Screenshot 2025-04-02 at 5 52 53 AM

Screenshot 2025-04-02 at 5 53 39 AM

adomasven:
Ok, fixed. Found some additional bugs in the bot detection and saving to server code along the way.

dstillman:
I'm still getting the error for the direct PDF save to server in Chrome on ScienceDirect. Expected?

adomasven:
Should be fixed

dstillman:
On https://forums.zotero.org/discussion/123152/feature-request-option-to-disable-automatic-markdown-in-notes, the snapshot is saved from the app rather than the extension.

PR #547, opened 2025-05-15
Move TranslatorTester support code here, update after rewrite

AbeJellinek:
Updates to support the [rewritten](https://github.com/zotero/translate/commit/a180ed48197fdb1e70f75ad58492def0a034b832) TranslatorTester.

`testTranslators.{js,css}` have been moved from zotero/translate with fairly minimal changes, but I replaced everything else because the previous implementation was [interesting](https://github.com/zotero/zotero-connectors/blob/9498491086c70b21a1d46ef33b46ea1dcd94c47d/src/common/tools/testTranslators/translatorTester_global.js#L40-L43).

There are now three test statuses (pending, success, failure) because the fourth ("unknown") didn't really make sense.


AbeJellinek:
(This is a lot of code so definitely no need for a comprehensive review @adomasven)

src/common/tools/testTranslators/testTranslators.html
```diff
@@ -27,11 +27,10 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
 	<head>
 		<script type="text/javascript" src="/browser-polyfill.js"></script>
-		<script type="text/javascript" src="/node_modules.js"></script>
 		<script type="text/javascript" src="/zotero_config.js"></script>
 		<script type="text/javascript" src="/zotero.js"></script>
 		<script type="text/javascript" src="/translate/promise.js"></script>
-		<script type="text/javascript" src="/translate/resource/zoteroTypeSchemaData.js"></script>
+		<script type="text/javascript" src="/utilities/resource/zoteroTypeSchemaData.js"></script>
```
adomasven, line 33: I've checked out zotero-schema directly into this repo now. You can load it at `/zotero-schema/zoteroTypeSchemaData.js`. Now when building the connector there's no need for a recursive submodule checkout.

adomasven:
If I go into Zotero Preferences -> Advanced and open the translator tester, it takes me to it, but it fails to load. Console logs:

```js
GET chrome-extension://apbjlgcgfpbcdeebnhfjaeogpccpejbc/tools/testTranslators/test.mjs net::ERR_FILE_NOT_FOUND
testTranslators.js:556 Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: chrome-extension://apbjlgcgfpbcdeebnhfjaeogpccpejbc/tools/testTranslators/translatorTester.mjs
```

src/common/tools/testTranslators/testTranslators.html
dstillman: Do you mean `/zotero-schema/schema.json`?

dstillman: But also, isn't this about the build files, not the source files? `/zotero-schema` doesn't exist in the build, as far as I can see. Isn't this data just in `/schema.js`?

adomasven: Ah yes, you're right. It is just `/schema.js`

AbeJellinek:
@adomasven: Did you update submodules? Need to pull latest zotero/translate. If that doesn't do it, try cleaning the build directory first?

AbeJellinek:
Never mind, same issue, on it

PR #561, opened 2025-07-22
Fix X-Metadata conversion error for non-ASCII attachment titles

aGolduck:
This commit implements a temporary workaround for handling non-ASCII characters in attachment titles when converting X-metadata through HTTP headers. The error occurs because HTTP headers only support ASCII characters (values <= 255), but the current implementation attempts to include UTF-8 characters directly.

Changes:
- Skip error reporting when attachment titles contain non-ASCII chars
- Preserve functionality while making the system more tolerant

Known issues:
- The current solution makes non-ASCII titles unreadable
- HTTP headers are not the appropriate mechanism for UTF-8 content
- A proper fix would require either:
  * Server-side decoding of metadata
  * A different protocol for metadata transmission

Error details:
[JavaScript Error: "XMLHttpRequest.setRequestHeader: Cannot convert argument 2 to ByteString because the character at index 162 has value 34593 which is greater than 255.
  Zotero.HTTP</this._requestXHR@moz-extension://bc5519f5-aa89-4cbb-840d-406e1c5063fd/http.js:243:12
  Zotero.HTTP</this.request@moz-extension://bc5519f5-aa89-4cbb-840d-406e1c5063fd/http.js:105:30
  Zotero.Connector</this.callMethod@moz-extension://bc5519f5-aa89-4cbb-840d-406e1c5063fd/connector.js:166:34
  Zotero.ItemSaver.saveAttachmentToZotero@moz-extension://bc5519f5-aa89-4cbb-840d-406e1c5063fd/itemSaver_background.js:58:26
  async*Zotero.Messaging</this.receiveMessage@moz-extension://bc5519f5-aa89-4cbb-840d-406e1c5063fd/messaging.js:84:20
  async*Zotero.Messaging</this.init/<@moz-extension://bc5519f5-aa89-4cbb-840d-406e1c5063fd/messaging.js:140:29
  " {file: "[object Object]"}]

### DIFF
--- a/src/common/itemSaver_background.js
+++ b/src/common/itemSaver_background.js
@@ -47,12 +47,14 @@ Zotero.ItemSaver.saveAttachmentToZotero = async function(attachment, sessionID, 
 		arrayBuffer = await this._fetchAttachment(attachment, tab);
 	}
 	
+	// Encode only the title to handle non-ASCII characters
+	const encodedTitle = btoa(unescape(encodeURIComponent(attachment.title || "Untitled Attachment")));
 	const metadata = {
 		id: attachment.id,
 		url: attachment.url,
 		contentType: attachment.mimeType,
 		parentItemID: attachment.parentItem,
-		title: attachment.title,
+		title: encodedTitle,
 	}
 
 	return Zotero.Connector.callMethod({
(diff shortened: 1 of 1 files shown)


dstillman:
Can you provide an example URL where you can reproduce the problem you're fixing?

adomasven:
We cannot take this PR, as this encodes all attachment titles as base64. In the meantime we recommend to change the translator to not use non-ASCII characters in title names. We will address it with #562.

Issue #567, opened 2025-08-04
Snapshots are attached to items saved from multiple-page


adomasven:
Report: https://forums.zotero.org/discussion/125873/problem-saving-snapshots-when-multiple-items-are-selected

This is currently disabled completely because we used to take snapshots from Zotero and they were more commonly wrong/paywalled, etc, but seems to be happening anyway with some level of bugginess to add.

PR #568, opened 2025-08-07
Fix snapshot attached to one of many saved items

abaevbog:
Without this, `ItemSaver._itemType` is not set, which breaks the temp workaround to avoid saving snapshots during multi-item checks (which relies on `this._itemType` == "multiple").

https://github.com/zotero/zotero-connectors/blob/7abdfdcc4dc07ae7e771362a10be6304b494c7dd/src/common/itemSaver.js#L158-L161

Fixes: #567

---

With this, if I save multiple items from https://www.metmuseum.org/art/collection/search?department=3&showOnly=highlights (link form the forums post), I no longer end up having an extra snapshot attached to one of the items. I do get some `attachment.mimeType is undefined` errors which stops proper attachments from being saved. It wasn't mentioned in the forums post though, so I wonder if it's just me?

### DIFF
--- a/src/common/inject/pageSaving.js
+++ b/src/common/inject/pageSaving.js
@@ -305,7 +305,8 @@ let PageSaving = {
 		}
 		items = this._processNote(items);
 		this.sessionDetails.items = items;
-		let itemSaver = new ItemSaver({sessionID, baseURI: document.location.href, proxy });
+		let itemType = translators[0].itemType;
+		let itemSaver = new ItemSaver({ sessionID, itemType, baseURI: document.location.href, proxy });
 		this.sessionDetails.itemSaver = itemSaver;
 		return itemSaver.saveItems(items, PageSaving._onAttachmentProgress, onItemsSaved)
 	},


abaevbog:
>  I do get some attachment.mimeType is undefined errors which stops proper attachments from being saved. It wasn't mentioned in the forums post though, so I wonder if it's just me?

I think this is just due to the fact that after translation, attachments objects do not have the `mimeType` set, which runs into issues here:

https://github.com/zotero/zotero-connectors/blob/7abdfdcc4dc07ae7e771362a10be6304b494c7dd/src/common/itemSaver_background.js#L239-L241

So is it the translator that needs to be fixed? I haven't done much with translators in general but from what I can see, `mimeType` is indeed [missing](https://github.com/zotero/translators/blob/fa5940aad8afa974ebeb2604c8341bbd9292d90b/The%20Met.js#L129-L132). At the same time, we do have the url that ends with `.jpg`, so is it worth to try to guess the mimeType based on that in the connector?

abaevbog:
> Yeah, I think if translators don't provide a mimeType then we accept whatever we get at the URL. I've pushed a commit with a fix.

Sounds good! I noticed that even though we accept all downloads now, the issue is that Zotero doesn't always know how to properly handle the attachment without the appropriate `contentType` for it. In this example here, the JPG attachment will be saved but the preview for it will not work in the itemPane. It makes the saved attachment not particularly helpful, and it's not very clear in the UI how to fix it. So I added another PR (linked above) to try to guess attachment mimeType for `png`, `jpeg`, and `pdf` based on the URL if one is not provided.

Obviously, all of this _should_ come from translators. But I thought that if it doesn't happen and we do allow an attachment without a provided mimeType to be saved, might as well do our best to save it right.

PR #569, opened 2025-08-11
Try to guess attachment MIME type if not provided by the translator

abaevbog:
If the attachment MIME type is not provided by the translator, check the URL to determine if it should be "image/jpeg", "image/png", or "application/pdf". This makes it more likely that such an attachment will be saved with the correct attachmentContentType, which is used to determine how the file is previewed in the itemPane.

Followup to https://github.com/zotero/zotero-connectors/pull/568

---

For comparison, saving an item from https://www.metmuseum.org/art/collection/search/327409 currently cannot be previewed, and is not very helpful:
 
Screenshot 2025-08-11 at 10 44 44 AM

With this change, the item will be saved with the jpeg contentType and the preview works as expected:

Screenshot 2025-08-11 at 10 45 14 AM

### DIFF
--- a/src/common/itemSaver_background.js
+++ b/src/common/itemSaver_background.js
@@ -234,8 +234,7 @@ Zotero.ItemSaver._fetchAttachment = async function(attachment, tab, attemptBotPr
 	try {
 		let xhr = await Zotero.HTTP.request("GET", attachment.url, options);
 		let result = Zotero.Utilities.Connector.getContentTypeFromXHR(xhr);
-		// extract mimeType from contentType header by stripping the charset (if any)
-		receivedMimeType = result.contentType.split(';')[0]?.trim();
+		receivedMimeType = result.contentType;
 
 		// If the attachment doesn't specify the mimeType, we accept whatever mimeType we got here.
 		// If translators want to enforce that a PDF is saved, then they should specify that!


adomasven:
Shouldn't we just use the Content-Type header returned by the server, and fallback to extension if that's unavailable?

abaevbog:
That's a great point actually! I moved things around a bit and re-pushed. If an attachment does not have the mime type, we'll set it to be the `contentType` from the request if it is provided. If it is not provided, we will try to guess it based on the url. That should cover pretty much all reasonable cases, I think.

I also moved `guessAttachmentMimeType` in to `utilities.js` to try to keep the files with the saving logic cleaner.

adomasven:
Can we add a test?

src/common/itemSaver_background.js
```diff
@@ -230,15 +230,26 @@ Zotero.ItemSaver._fetchAttachment = async function(attachment, tab, attemptBotPr
...
 	// Bot bypass is not supported in Safari (cannot intercept file download popup)
 	attemptBotProtectionBypass = attemptBotProtectionBypass && !Zotero.isSafari;
-	var contentType;
+	var receivedMimeType;
 	try {
 		let xhr = await Zotero.HTTP.request("GET", attachment.url, options);
 		let result = Zotero.Utilities.Connector.getContentTypeFromXHR(xhr);
-		contentType = result.contentType;
+		// extract mimeType from contentType header by stripping the charset (if any)
+		receivedMimeType = result.contentType.split(';')[0]?.trim();
```
adomasven, line 238: Sorry, the diff before didn't include the lines above and I didn't notice we were already getting the content type from the response headers in the `getContentTypeFromXHR()`, which also extracts the charset separately and returns it. As such, this additional logic splitting the `;` isn't necessary.

I think we just need to check if attachment has a mime type, and if it doesn't, set it to `result.contentType`

PR #572, opened 2025-08-20
Add support for pushing multiple tabs to zotero

odneill:
Add a "Save to Zotero" option to the Tab context menu, such that the connector attempts to save all currently selected tabs to the library in one go, avoiding having to manually click connector on each tab.


adomasven:
Thanks for the addition! I had to clean up the code a little to make it conform to existing context menu code.

This feature only works on Firefox, tab context menus are not supported on Chromium browsers.

PR #576, opened 2025-09-15
Add OpenAthens support

### LINKED ISSUE #571: Better support for OpenAthens proxy
OpenAthens isn't a traditional proxy in that they don't rewrite the URL. There is no way to know whether the user is authenticated for a given resource. The only reliable thing to check is redirects via the openathens redirector. Thankfully, those follow a flow via standardized urls.

- Detect OpenAthens proxy via the redirector
- Add domains to associated list that we detect being redirected via the redirector
- Automatically redirect via the redirector at least once per browser session (and perhaps every 6-24 hours, configurable?) for the associated hostnames

We should also handle users moving to different institutions, or institutions switching to a different proxy provider (EZProxy <-> OpenAthens), and if we detect hostname overlap, we should disable the other proxy.

adomasven:
Closes #571

As per the issue, implements OpenAthens proxy support, by checking for navigation via the redirector. A proxy is automatically added and will be identified by `https://go.openathens.net/redirector/[PROXY_URL]?url=`. Then when navigating to associated hostnames, we auto-redirect via the redirector every 12 hours.

Added a switch to proxy settings that allows to switch between EZProxy and OpenAthens proxies. Also some misc CSS updates and refactoring in the preferences.

I've also changed the logic of proxies, such that if there are multiple proxies, we will automatically steal hostnames from one proxy to another when a hostname is detected to be used with one of the proxies.

I'll also have a PR after this is merged that fully localizes preferences page and the rest of the strings in the connector (but it depends on the changes here).

### DIFF
--- a/src/common/messages.js
+++ b/src/common/messages.js
@@ -280,6 +280,7 @@ var MESSAGES = {
 	},
 	Proxies: {
 		loadPrefs: false,
+		validate: true,
 		save: false,
 		remove: false,
 		toggleRedirectLoopPrevention: false
--- a/src/common/preferences/preferences.css
+++ b/src/common/preferences/preferences.css
@@ -1,5 +1,5 @@
 body {
-	font-family: Lucida Grande, Tahoma, Arial, sans;
+	font-family: Lucida Grande, Tahoma, Arial, sans-serif;
 	font-size: 12px;
 	background-color: #b7bfcb;
 }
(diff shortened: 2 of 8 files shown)


adomasven:
We could name it "URL-rewriting proxy", but presumably 99% of the people are going to be on EZProxy, so I'd rather the 1% with different service providers be confused.

PR #609, opened 2026-03-15
Improve cross-origin request handling in http.js

tobiasdiez:
The method `_requestFetch` is only called if `isManifestV3 = true`, so the body of the method can be simplified a bit (no proxy + parser needed)

May I ask why in MV3 all http requests are performed via the background script, while the old `_requestXHR` method performs the request from the content script if it's not a cross-domain request (otherwise it's also done via the background)? I'm asking as I have problems with some requests that don't seem to work in the background script [with MV3 but on Firefox], but work well from the content script.

Finally, I wanted to mention that the current behavior to allow the content script to pass the requested url seems to be quite insecure (see https://developer.chrome.com/docs/extensions/develop/concepts/network-requests#xhr-vs-content-scripts). An attacker could inject a malicious content script/webpage script that sends the `co-http` request to the zotero background page and thereby perform arbitrary cross-domain requests.

### DIFF
--- a/src/common/http.js
+++ b/src/common/http.js
@@ -281,33 +281,17 @@ Zotero.HTTP = new function() {
 	
 	this._requestFetch = async function(method, url, options) {
 		if (Zotero.isInject) {
-			// Make a cross-origin request via the background page, parsing the responseText with
-			// DOMParser and returning a Proxy with 'response' set to the parsed document
-			let isDocRequest = options.responseType == 'document';
-			let coOptions = Object.assign({}, options);
-			if (isDocRequest) {
+			// Make a cross-origin request via the background page
+			const coOptions = Object.assign({}, options);
+			if (options.responseType == 'document') {
 				coOptions.responseType = 'text';
 			}
 			if (Zotero.isSafari && options.headers['User-Agent']) {
 				coOptions.headers['Cookie'] = document.cookie;
 			}
-			return Zotero.COHTTP.request(method, url, coOptions).then(function (xmlhttp) {
-				if (!isDocRequest || Zotero.isManifestV3) {
-					xmlhttp.responseType = options.responseType;
-					return xmlhttp;
-				}
-				
-				Zotero.debug("Parsing cross-origin response for " + url);
-				let parser = new DOMParser();
-				let contentType = xmlhttp.getResponseHeader("Content-Type");
-				let doc = parser.parseFromString(xmlhttp.responseText, Zotero.HTTP.determineDOMParserContentType(contentType));
-				
-				return new Proxy(xmlhttp, {
-					get: function (target, name) {
-						return name == 'response' ? doc : target[name];
-					}
-				});
-			});
+			const xmlhttp = await Zotero.COHTTP.request(method, url, coOptions);
+			xmlhttp.responseType = options.responseType;
+			return xmlhttp;
 		}
 		
 		try {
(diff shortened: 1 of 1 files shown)


tobiasdiez:
Just as a disclaimer: neither the code changes nor the PR description were written by an LLM. All mistakes are mine alone (and it's likely that I made some as I'm still trying to understand most of the codebase). Being an OSS maintainer myself, I'm used to the LLM swamp, thus can understand your feelings about it and try not to contribute to it myself.

> > May I ask why in MV3 all http requests are performed via the background script, while the old _requestXHR method performs the request from the content script if it's not a cross-domain request (otherwise it's also done via the background)?
> 
> Because the content XHR/fetch can be limited in various ways, such as via CSP, HSTS, and others, which breaks requests and subsequently page translation. XHR is used in MV2, because you can manually set referrer on it, unlike with fetch, and some translators do not work without that. On Chromium browsers we inject referrer headers via DNR. We could maybe do that on Firefox too and only use fetch, but I believe I've tried that before at some point and Firefox actually blocks that as a security policy. Setting referrer via XHR only sends the origin though, whereas for same-origin requests the full page url is sent, which is required for APA PsychNet and possibly other translators, which is why that code path exists.

Thanks, this is very helpful! I'll see if I can make fetch work on Firefox as well.

> > Finally, I wanted to mention that the current behavior to allow the content script to pass the requested url seems to be quite insecure (see https://developer.chrome.com/docs/extensions/develop/concepts/network-requests#xhr-vs-content-scripts). An attacker could inject a malicious content script/webpage script that sends the `co-http` request to the zotero background page and thereby perform arbitrary cross-domain requests.
> 
> This seems like an LLM hallucination. Extension content scripts are isolated from webpage scripts and other extensions scripts.

I'm far from an expert on these matters but to cite from the link above:
> In the approach above, the content script can ask the extension [background script] to fetch any URL that the extension has access to. A malicious web page may be able to forge such messages and trick the extension into giving access to cross-origin resources. Instead, design message handlers that limit the resources that can be fetched.

> 
> > The method `_requestFetch` is only called if `isManifestV3 = true`, so the body of the method can be simplified a bit (no proxy + parser needed)
> 
> This seems like an LLM hallucination. These changes would break HTTP requests for documents from content scripts.
 
The only call to `_requestFetch` that I could find was in https://github.com/zotero/zotero-connectors/blob/92d2c56de53c82526bc68bab378b014dcc3b007a/src/common/http.js#L101-L106, which does limit it `isManifestV3 = true`. Are there other calls? And if `isManifestV3` is true in `_requestFetch`, this if condition is always true and the method exists early without using the DOMParser:
https://github.com/zotero/zotero-connectors/blob/92d2c56de53c82526bc68bab378b014dcc3b007a/src/common/http.js#L295-L298

> > I'm asking as I have problems with some requests that don't seem to work in the background script [with MV3 but on Firefox], but work well from the content script.
> 
> If you have development questions, we provide support in the [zotero-dev](https://groups.google.com/forum/#!forum/zotero-dev) mailing list. It would be most productive if you explained what you're trying to achieve in your own words, instead of relying on LLM and the incorrect assumptions they make without extensive knowledge of the codebase and the various edge cases it's intended to handle.

Thanks, at the moment I'm still trying to understand the code better / why it's not working in my context. But what I'm roughly trying to archive at the moment is to use MV3 in Firefox as well.

PR #630, opened 2026-07-09
Safari web extension

adomasven:
Switches the Safari extension out from the old app extension to the web extension framework.

Benefits:
- Proxy detection and redirection sort of works sometimes
- CSL import from previously-defined URLs works (zotero.org/styles, github.com)
- We can check whether the user has granted permissions to run on all websites, and show a popup if they haven't.

Missing Safari Web Extension features:
1. DNR rule conditions for response headers
2. Retrieving cookies for any url/domain (which seems like an undocumented bug-ish behavior, since the API to do this technically exists, but effective no-ops).

As a result, things that are not improved over app extension:
- Attachment saving that requires access to browser-stored cookies for specific URLs (2)
- Bot protection bypass (cf turnstile, sciencedirect) for attachment saving (2)
- CSL and RIS/RDF/BibTeX import from not previously defined URLs (1)

We could maybe sort of in theory make import from not previously defined URLs work by observing the content-type via `onHeadersReceived` which doesn't block, but when I tried to get this to work, I had safari report a tabId that was invalid, and when I tried to work around that, it froze the entire tab, without (reliably) getting to the import confirm dialog.

There are some minor and some egregious bugs in Safari web extension implementation:
- `webRequest.onHeadersReceived` is often not sent, especially when navigating from new tab
- `webRequest.onHeadersReceived` reports old url after 302 redirect. This and the above causes proxy detection to be inconsistent and to fail more often than succeed.
- i18n implementation is buggy and we have to roll our own
- Message passing via tabs.sendMessage fails to extension injected pages, which requires a fairly annoying workaround path via postMessage
- Extension files must have the UTF-8 BOM, otherways non-ASCII characters in code are garbled (an additional build step)

-----------

I tested this using Safari Preferences -> Developer -> Add Temporary Extension... I didn't use the xcode project as it's not needed/doesn't introduce unnecessary delay.

Safari actually has a keyboard binding to open the background page of an extension for debugging, but when I use it the opened devtools don't load properly, and it only works if I open them when clicking with the cursor, and the background devtools close after every extension reload. Unfortunate.

All the relevant codepaths that I could think of, such as saving, saving via proxy, snapshots, google docs, proxy reloading, import were tested and work as expected aside from exceptions above.


dstillman:
Take a look at b3dfc6a1ad03e16bef85dadd4c69f04e9f48125b. That hopefully limits proxy redirection failures to those from a new tab?

dstillman:
Added a workaround for SPA navigation that fixes translator detection on, e.g., YouTube. (We can fold that into the earlier commit that feature-detected it.)

adomasven:
I've removed the safari document.cookie fallback (instead of clarifying the comment) since with the cookies workaround it's not needed.

dstillman:
Well, I left it in because there are at least a couple situations where the Cookie API doesn't work:

1. Before Safari 18, getAll() had documented limitations (default store only, no HttpOnly): https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/getAll#browser_compatibility — Safari 18 goes back to Ventura (13), so this is only macOS Monterey (12) and earlier, so not a big deal, but also no downside.
2. [`cookies.get() and cookies.getAll() have empty results on first launch`](https://bugs.webkit.org/show_bug.cgi?id=315310) — current bug, PR just landed

But not a big deal if we want to get rid of it.

dstillman:
Merged

PR #631, opened 2026-07-28
Add host permission prompts for Safari

adomasven:
- Upon first enabling the Zotero Connector to run on any page, prompt about required permissions for localhost, and suggested permissions for all websites
- Once per browser start if localhost permissions are missing prompt about them, and trigger a ping to Zotero, which will show a Safari permission dialog (only if Connector is allowed to run on a given page)
- If attempting to save without required permissions, prompt
- If localhost permissions are enabled, but Zotero is offline, prompt about repo.zotero.org permissions required for translators
- If saving with Zotero offline, prompt about api.zotero.org permissions required for saving to web library
- If https://*/* permissions are missing, prompt about it every time the user opens the preference pane

-------

Sometimes the individual (or ALL) host permissions may be configured to deny, but will proceed anyway when requested via background page. This is even if the browser's own API will report the permission for a given host as missing. Sometimes it persists with Safari restarts, other times it clears with merely reloading the extension. As such, sometimes we are showing prompts for permissions for e.g. api.zotero.org, and then those requests succeed anyway.

We include the instructions on how to enable permissions for a given host in every prompt, since we don't know if a host we don't have permissions for is set to "Ask" or "Deny". If it's "Ask", Safari will follow our prompt by displaying its own native one, where the user can allow Zotero Connector the host permission. If it's "Deny", no such prompt is shown.


dstillman:
OK, made a bunch of tweaks to the dialogs, and did a huge amount of testing to try to get the prompt experience right in Safari with different combinations of initial permissions. I'm going to merge to get this into a beta before Z10, but let me know if anything looks wrong.

dstillman:
Merged.

Re: the gdocs submodule commit for repo.zotero.org/settings, we could probably just remove that request instead now that we're always using v2.

src/safari/reinjectGuard.js
```diff
@@ -0,0 +1,65 @@
...
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+(function() {
+// Safari injects the extension's content scripts into already-loaded matching pages when the
+// user grants new host permissions. The scripts run in the same content world as the first
+// copy, which recreates the Zotero namespace and throws on top-level let and const
```
adomasven, line 29: I don't understand this. If the host has no permission  to run on the page, there is no "first copy", because nothing gets injected until permission is granted. Is this for multiple tabs with the same host? Or did you observe this for a single tab?

In my testing I didn't see any duplicate injection behavior, but this explanation doesn't make sense to me.

dstillman: This was trying to get Google Docs to work on first run. If no host permissions, you have a Z (previously gray webpage icon) on the save button. If you click that and grant access only to docs.google.com but not 127.0.0.1, we then show an alert about 127.0.0.1. After you click "OK", we trigger the Safari access prompt for 127.0.0.1. After granting access, the new Zotero menu in Google Docs wouldn't work, and this (including a couple extra debugging lines) showed up in the console:

Screenshot 2026-08-07 at 2 46 44 PM

(Initially two Zotero menus would show, because it was injecting each time, but we now prevent that by checking for the DOM element.)

src/browserExt/background.js
```diff
@@ -1052,11 +1053,26 @@ Zotero.Connector_Browser = new function() {
...
+	 * site access, so inject them on demand before performing a user action, and give
+	 * translator detection a moment to report before a save mode is chosen
+	 */
+	async function _ensureScriptsInjected(tab) {
+		if (!Zotero.isSafari) return;
+		await Zotero.Connector_Browser.injectTranslationScripts(tab);
+		let tabInfo = Zotero.Connector_Browser.getTabInfo(tab.id);
+		for (let i = 0; i < 30 && !tabInfo.translators; i++) {
+			await Zotero.Promise.delay(100);
+		}
```
adomasven, line 1067: If a page has no translators, doesn't this cause a 3 second delay every save?

dstillman: No, `tabInfo.translators` with no detected translators is an empty array, so it's not falsy.

But if the user navigates during the loop, this wouldn't get the new value, which might result in a stall, so I'll change this to get the tab info again on each loop.

src/safari/reinjectGuard.js
adomasven: I mean google docs menu injection depends on Zotero scripts being loaded before their editor finishes initializing, so for google docs permission granting it would make sense to reload the page as an exception. But I've not seen duplicate script injection or that causing issues for translation in other places.

Either way, this comment seems incorrect, overly-verbose, and likely to further trip up LLMs and people in what goes on here, or for what reasons.

src/browserExt/background.js
dstillman: https://github.com/zotero/zotero-connectors/commit/8c58203c80255fabf6e64660ac31ad3ba9c8f321

src/safari/reinjectGuard.js
dstillman: Easy to see the duplicate injection on any site:

1. Set all access permissions for the temp extension to Ask.
2. Uninstall the extension.
3. Restart Safari.
4. Load forums.zotero.org, show the page console, and enable Persist Logs.
5. Add the temporary extension and reload the forums tab.
6. Click the Z and grant access to forums.zotero.org for one day.
7. Get the 127.0.0.1 warning. Click OK.
8. In the Safari access prompt for 127.0.0.1, enable "Remember for other websites" and click one of the allow options. It doesn't happen if you don't enable "Remember for other websites".

Screenshot 2026-08-10 at 1 51 19 PM

I didn't test all the ways it might break things on non–Google Docs pages, but it almost certainly would break something, and it doesn't seem like a state we want to operate in, so reloading is the safest option.

I've clarified the comment in https://github.com/zotero/zotero-connectors/commit/c5a97be7ce4a86548f0c9ccbeeec2f2bba6668cc to note that it depends on "Remember for other websites".

PR #632, opened 2026-08-04
Harden the translate runtime environment

adomasven:
Implement the following security guarantee:
> Translator detection has no supported direct or privileged HTTP
> capability and runs in a disposable, role-bound realm.

This ensures that if the user environment for some reason ends
up including translators that attempt to monitor and exfiltrate user
browsing data via the detect() method which runs on page load, it is no
longer possible.

This is implemented by moving translation on all platforms to background
(Chromium MV3: offscreen page) based translate frames. Each detect and translate
call creates a new frame in the background, which runs the translator
code in an isolated sandboxed environment. Translator detect calls run
in an environment that don't have any way to perform HTTP calls.

@AbeJellinek Could you check if this breaks translator CI in the repo? Also feel free to review this and/or ask questions.


AbeJellinek:
> Translator detect calls run in an environment that don't have any way to perform HTTP calls.

Will this break unAPI?

Sounds like a step in the right direction, though. I'll take a look.

AbeJellinek:
This does break unAPI, so we'll need to figure out what to do about that before merging. (Maybe not much? Just make unAPI detection return a generic type upon seeing a link rather than trying to resolve it?) Other than that, though, CI works fine and I think this is good.

adomasven:
@dstillman This is ready now

PR #634, opened 2026-08-12
Localize the rest of strings

### LINKED ISSUE #58: Localize connectors

adomasven:
Localize the rest of the strings. Closes #58

Localize preferences, browser actions and menus, import/save
dialogs, progress errors, proxy prompts, version warnings,
and extension metadata.

Add contextual accessible names for proxy controls

Preserve Safari fallback behavior for newly added strings.

### DIFF
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -261,12 +261,11 @@ Zotero.Connector_Browser = new function() {
 	 */
 	this.onIncompatibleStandaloneVersion = function(connectorVersion, clientVersion) {
 		if (_incompatibleVersionMessageShown) return;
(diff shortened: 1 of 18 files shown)


abaevbog:
Looks great to me!


The only thing I may suggest is if we are adding accessible tweaks to proxy controls here, might as well link the labels to `Preferences-Proxies-hostSelect` and `Preferences-Proxies-proxySelect`? Now when those are focused they're announced as e.g. "empty listbox". Adding a `for=` attribute to the `label`s above pointing at those nodes' ids should attach to those `<select>` nodes a proper accessible label.

PR #635, opened 2026-08-13
Fix attachment cookie handling with Firefox First-Party Isolation

### LINKED ISSUE #226: Can't get cookies in Firefox first-party-isolation mode
When first-party isolation is enabled in Firefox, browser.cookies.getAll() will fail if `firstPartyDomain` isn't provided, causing all saves to fail. This is all well and good.

However, [according to MDN](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/cookies/getAll), when Firefox is in first-party-isolation mode, passing `firstPartyDomain: null` to `browser.cookies.getAll()` should cause all cookies to be returned:

> This property must be supplied if the browser has first-party isolation enabled. You can however pass null in this situation. If you do this, then cookies with any value for firstPartyDomain, as well as cookies which do not have firstPartyDomain set at all, will be included in the results.

As of Fx60.0b7, though, that doesn't seem to be working, returning no cookies instead.

In 60.0b7 it does work to set the domain explicitly (e.g., 'gmu.edu'), but we can't get that correctly without the public suffix list, which [isn't yet available to WebExtensions](https://bugzilla.mozilla.org/show_bug.cgi?id=1315558).

We could try to pass an explicit domain for known eTLDs, something like:
```javascript
let domain = null;
try {
	let url = require('url');
	let hostname = url.parse(tab.url).hostname;
	let matches = hostname.match(/([^.]+\.(?:[^.]+|ac\.[a-z]{2}|edu\.ar|co\.uk))$/);
	if (matches) {
		domain = matches[1];
	}
}
catch (e) {
	Zotero.debug(e, 2);
}
```
…but that would risk breaking things when FPI wasn't enabled, so I don't think it's worth it.

So for now, we just pass `null`, which will allow saves to go through, even if attachments that rely on cookies will fail.

I'm not sure what the best fix here will be (and how using a working `null` would affect privacy when those cookies were passed to Zotero), but at the moment I'm not sure we can do better than this.

pellaeon:
This PR:

1. Resolve cookies using the originating tab’s Firefox FPI and container context, and avoid combining cookies from unrelated partitions.

2. Gracefully continue attachment downloads when cookie enumeration fails, scope temporary Cookie headers to the exact request URL, and preserve browser cookie semantics when injecting contextual cookies.

3. Also retain compatibility with older Chromium versions that reject partitionKey and preserve the tab context during bot-bypass retries.

Fixes #226 

The code changes are proposed by a LLM and manually reviewed by me line-by-line. Manual testing shows that this PR fixes PDF saving issues on Firefox with FPI enabled.

### DIFF
--- a/src/browserExt/background.js
+++ b/src/browserExt/background.js
@@ -109,21 +109,72 @@ Zotero.Connector_Browser = new function() {
 	}
 
 	/**
-	 * Gets cookies from the store associated with a tab. In Safari, the default cookie store
-	 * may differ from the store used by browser tabs, so it must be resolved at runtime. If
-	 * no tab is provided, the active tab in the current window is used.
+	 * Gets cookies from the store associated with a tab. Firefox container tabs and Safari tabs
+	 * may use a different store from the extension default, so it must be resolved at runtime.
+	 * In Safari, if no tab is provided, the active tab in the current window is used.
 	 *
 	 * @param {Object} details
-	 * @param {Number} tabId
+	 * @param {Number|Object} tabOrId A snapshot of the originating tab is preferred so that a
+	 * navigation while saving cannot change the cookie context.
 	 * @return {Promise<browser.cookies.Cookie[]>}
 	 */
-	this.getAllCookies = async function(details, tabId=null) {
+	this.getAllCookies = async function(details, tabOrId=null) {
 		details = {...details};
+		let tab = typeof tabOrId === 'object' ? tabOrId : null;
+		let tabId = tab?.id ?? tabOrId;
+		if (Zotero.isFirefox && tabId != null && !tab) {
+			tab = await browser.tabs.get(tabId);
+		}
+		if (Zotero.isFirefox && !details.storeId) {
+			if (tab?.cookieStoreId) {
+				details.storeId = tab.cookieStoreId;
+			}
+		}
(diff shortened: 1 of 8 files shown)


adomasven:
Could you provide URLs, procedures and software versions with which you manually tested this and confirmed it works?

pellaeon:
* PDF URL: https://www.usenix.org/system/files/conference/usenixsecurity26/sec26_prepub_hu.pdf
* Firefox version: 151.0.3 (amd64, Ubuntu 26.04, from https://launchpad.net/~mozillateam/+archive/ubuntu/ppa/+packages )
* Firefox `about:config` : set `privacy.firstparty.isolate` to `true`

Before fix behavior:
1. navigate firefox to PDF URL
2. open Zotero app (flatpak 9.0.4)
3. click the Zotero connector icon to save
4. the icon turns into a red X mark, says "save failed" , PDF did not show up in Zotero app

After fix behavior:
4. the icon turns into a green check mark, PDF shows up in Zotero app.

adomasven:
Thanks for the PR. This wasn't quite in the state that we'd be able to merge, but I used it as a starting point for a smaller change. It's now available in the <a href="https://www.zotero.org/support/dev_builds">Zotero Connector Beta</a>

PR #639, opened 2026-09-17
Let users dismiss the Safari all-websites recommendation

dstillman:
We currently say that all-websites access is required once per browser session whenever it hasn't been granted. Site access in Safari is technically optional, so call it a "recommendation" and add a "Don't Show Again" checkbox.

The checkbox appears only when the prompt is purely a recommendation. A prompt that also covers required access, such as missing localhost access, reappears until that access is granted.

https://forums.zotero.org/discussion/133713/safari-pop-up-opens-every-time-safari-opens


adomasven:
I've changed the logic so that we show the recommendation to enable all hosts if we're showing the prompt anyway due to a missing permission for some specific domain (e.g. missing 127.0.0.1). The don't show again checkbox only disables prompts that would appear only to recommend without requesting specific domain access.

Good to squash/merge.

